PR #23, opened 2018-11-02
Added default HTTPs site.

jlesage:
Without a default site for HTTPs, another configured host will be wrongly used.

For example, when trying to access, via HTTP, a host that has not been setup yet, we get the "Congratulations" page as expected.  However, doing the same via HTTPs does't show this default page, but shows a configured proxy host (for which SSL is enabled).

The same issue happen when configuring a host without enabling SSL:  Access the host over HTTP works fine, but accessing the same host via HTTPs will show the page associated to another configured host.

Since there is no certificate associated to the default site, its configuration only return the special code `444`, which makes nginx to close the connection.

### DIFF
--- a/rootfs/etc/nginx/conf.d/default.conf
+++ b/rootfs/etc/nginx/conf.d/default.conf
@@ -36,3 +36,17 @@ server {
     root /var/www/html;
   }
 }
+
+# Default 443 Host
+server {
+  listen 443 ssl default;
+  server_name localhost;
+
+  access_log /data/logs/default.log proxy;
+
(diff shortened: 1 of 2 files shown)


jlesage:
I need to adjust the change, as it seems to break proxy hosts with SSL...

jlesage:
Ok I adjusted the default site and now it's working fine.

jc21:
Great stuff :) I'll add a bit of debug output for the dummy cert generation prior to the next release.

Thanks!

PR #74, opened 2019-02-19
Feature/custom locations

kolbii:
I've been using this tool for making requests between the projects and services, it's a great tool for that. But in that case, if a project has 2+ location path it gets difficult to use and understand.
Please take a look at this PR. I've added a new pane to Proxy host dialog as "Custom locations" where you'd be able to manaeg 2+ location blocks

[screenshot1](https://imgur.com/3ffSp5y)
[screenshot2](https://i.imgur.com/gsFUaAc.png)

PR #75, opened 2019-02-19
#65 custom config of location / on proxy host pane

### LINKED ISSUE #65: Allow full nginx config editing
do you think we can have full access to to config file under advanced settings so we can edit as we like please

kolbii:
Fixes #65 
Adds a new field to Proxy host and puts the generator that code into the root location block's body.
[screenshot](https://imgur.com/a/eTIos3M)

### DIFF
--- a/src/backend/migrations/20190219223711_root_config.js
+++ b/src/backend/migrations/20190219223711_root_config.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const migrate_name = 'rootpath_config';
+const logger       = require('../logger').migrate;
+
+/**
+ * Migrate
+ *
+ * @see http://knexjs.org/#Schema
+ *
+ * @param   {Object}  knex
+ * @param   {Promise} Promise
+ * @returns {Promise}
+ */
+exports.up = function (knex/*, Promise*/) {
+    logger.info('[' + migrate_name + '] Migrating Up...');
+
+    return knex.schema.table('proxy_host', function (proxy_host) {
+            proxy_host.text('advanced_rootconfig').notNull().defaultTo('');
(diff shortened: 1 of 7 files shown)


jc21:
Docker Image for build 4 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-75`

kolbii:
Is this tag`jc21/nginx-proxy-manager:github-pr-75` available on DockerHub? Trying to get this image down but it looks like still unpublished

jc21:
Docker Image for build 5 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-75`

jc21:
Hmm yeah it said it uploaded it but I can't see it either.

jc21:
Docker Image for build 6 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-75`

jc21:
Thanks for contributing to fix this problem. I've been thinking about it and the problem has a simpler solution. The template render of the host nginx config file just needs to ignore the default location when one is present in the advanced config. I've made a quick prototype and it works well.

I'd like to keep the configuration simple, and by that I mean not confusing for those not educated in nginx internals. By adding a special configuration on top of the advanced configuration page seems a bit much.

I'll push another PR with my idea should you want to compare and comment.

kolbii:
You're right. It's simpler like having another input for overriding any params. I'm closing this PR.
However, what do you think about the custom locations section? I think it'd be easier to manage 2+ simple proxied location blocks like writing advanced config. Please take a look at the other one #74 by me.

PR #79, opened 2019-02-21
Ignore default location when defined in advanced config

### LINKED ISSUE #65: Allow full nginx config editing

jc21:
Fixes #65 by ignoring the built-in default location when it's detected in the advanced configuration.

In essence, the regex searches for `location / {` in many forms.

### DIFF
--- a/src/backend/internal/nginx.js
+++ b/src/backend/internal/nginx.js
@@ -1,5 +1,3 @@
-'use strict';
-
 const _          = require('lodash');
 const fs         = require('fs');
 const Liquid     = require('liquidjs');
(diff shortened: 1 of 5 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-79`

kolbii:
Good idea.
I use the config block a lot but I'm feeling bad while I'm working on configuring. (indent, highlight, etc)
Wouldn't [CodeMirror](https://codemirror.net/) be good stuff instead of simple textarea?

Issue #26, opened 2018-11-08
Add configurable error pages


jc21:
It would be nice to allow the user to specify the content for the following pages
- upstream 404
- upstream 500
- upstream unavailable error
- default page when no matching hostname is configured

Daxx13:
>     default page when no matching hostname is configured

+1 to this, will be a great enhancement.

didusee:
+1

ghallford:
I would like the option to have the default page when no hostname is specified return just a 404 error message. Or just make it so that if anything is hit that is not a valid hostname proxy, it returns a 404 including if the default page is hit from the client using IP of the server in the URL

tophee:
+1

aosmirnov:
> It would be nice to allow the user to specify the content for the following pages
> 
> * upstream 404
> * upstream 500
> * upstream unavailable error
> * default page when no matching hostname is configured

This feature is really needful!

Mohan-cloud:
@tophee Is the redirection to a public site still the best solution?

l4rm4nd:
@jc21 any updates or plans on this feature enhancement?

PrzemekSkw:
> +1
> 
> Sorry if this is a dumb question, but what is currently the best work around for not exposing that page saying "Congratulations! You've successfully started the Nginx Proxy Manager"? My quick solution was to create a redirect host for my bare IP and redirect to some public website. So Whoever visits my IP will just be bounced back into the internet... :-)

Hello, I also see that Congratulations page on my external IP. Can You tell how to make that redirection?
Regards.

aosmirnov:
> > +1
> > Sorry if this is a dumb question, but what is currently the best work around for not exposing that page saying "Congratulations! You've successfully started the Nginx Proxy Manager"? My quick solution was to create a redirect host for my bare IP and redirect to some public website. So Whoever visits my IP will just be bounced back into the internet... :-)
> 
> Hello, I also see that Congratulations page on my external IP. Can You tell how to make that redirection? Regards.

You should go to settings page in web UI and change Default site settings from Congratulations page to 404 page or Custom HTML or configure a redirection.

PrzemekSkw:
Hi, thanks for Your answer. I miss that settings tab. I thougt I have to make someting in command line :D
Thanks.

Anas-Sherif:
Any updates? I really need this.

b4sh1t1:
> > > +1
[...]

PR #91, opened 2019-03-04
Part of #26 - Default Site Customisation

jc21:


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-91`

PR #99, opened 2019-03-10
This commit resolves #98 so custom location can forward to custom path.

### LINKED ISSUE #98: URI support in proxy_pass
The nginx `proxy_pass` directive supports an URI:
```
proxy_pass http://localhost:8000/uri/;
```
See http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass

It would be useful to support this.  This could be especially useful for custom locations, where one would want to forward without the location part.  For example, with:
```
location /sub/ {
  proxy_pass http://192.168.1.1:8080/;
}
```
The following forwarding would occur:
`http://example.com/sub/index.html` -> `http://192.168.1.1:8080/index.html`

Or with:
```
location /sub/ {
  proxy_pass http://192.168.1.1:8080/other;
}
```
We would have:
`http://example.com/sub/index.html` -> `http://192.168.1.1:8080/other/index.html`

kolbii:
This PR resolves #98 sub-folder forwarding.

### DIFF
--- a/src/backend/internal/nginx.js
+++ b/src/backend/internal/nginx.js
@@ -151,7 +151,16 @@ const internalNginx = {
 
             const locationRendering = async () => {
                 for (let i = 0; i < host.locations.length; i++) {
-                    renderedLocations += await renderer.parseAndRender(template, host.locations[i]);
+                    let locationCopy = Object.assign({}, host.locations[i]);
+
+                    if (locationCopy.forward_host.indexOf('/') > -1) {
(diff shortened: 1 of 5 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-99-amd64`

PR #114, opened 2019-04-02
Fixed directory traversal vulnerability.

jlesage:
Management interface can leak local file system.  For example, accessing an url like this is allowed:

http://test.com:81/%2e%2e%2f%2e%2e%2f%2e%2e%2f%2e%2e%2f%2e%2e%2fetc/passwd

This PR fixes this by not allowing access to resources outside `dist/`.

### DIFF
--- a/src/backend/routes/main.js
+++ b/src/backend/routes/main.js
@@ -3,6 +3,7 @@
 const express = require('express');
 const fs      = require('fs');
 const PACKAGE = require('../../../package.json');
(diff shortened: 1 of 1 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-114-amd64`

PR #130, opened 2019-05-03
Update the docker-compose

skywalkerisnull:
Added the ports to be open on the jc21/mariadb-aria.

I was getting a connection error of: 
`app_1  | [5/3/2019] [2:04:54 AM] [Global   ] › ✖  error     connect ECONNREFUSED 192.168.2.109:3306` 

I was able to resolve this by adding in the ports that need to be opened in the docker-compose script.

### DIFF
--- a/doc/INSTALL.md
+++ b/doc/INSTALL.md
@@ -59,6 +59,8 @@ services:
   db:
     image: jc21/mariadb-aria
     restart: always
+    ports:
+      - 3306:3306
     environment:
       MYSQL_ROOT_PASSWORD: "password123"
       MYSQL_DATABASE: "nginxproxymanager"


jc21:
Technically this change to the example won't do anything to solve your problem. The example you're editing starts the containers in docker `bridge` mode and you're using it in `host` mode presumably, for the `nginx-proxy-manager` container but not the `mariadb-aria` container.

jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-130-amd64`

skywalkerisnull:
Yes you are correct, this machine has docker using host networks.

Just did a few tests to get it working without the explict port number in the docker compose file, I needed to change the host from local host IP to 'db' like what you have in the docs/example/config.json

brusezot:
i don't understand why ? jc21 can you explain with more detail your comment : 

"Technically this change to the example won't do anything to solve your problem. The example you're editing starts the containers in docker bridge mode and you're using it in host mode presumably, for the nginx-proxy-manager container but not the mariadb-aria container."

i would like resolve this problem please ;) 

thanks you for your job

jc21:
In the docker-compose.yml file that has been changed in this PR, there are 2 services: `app` and `db`.

There is no mention of [network_mode](https://docs.docker.com/compose/compose-file/#network_mode) on any service which means that the services are started using `bridge` mode. Bridge mode means that the services can talk to each other on "exposed" ports inside their own little NAT. In the case of these services, 3306 is exposed on the `db` service; this is done with the Dockerfile already, and is not required in the docker-compose configuration.

Your proposed change only exposes the `db` 3306 port to the **host** network, which the `app` service is not using, so your change is pointless in helping you connect the `app` to the `db`.

So my comment above was assuming either you strictly specified an IP of `192.168.2.109` as the database server in the NginxProxyManager configuration file (instead of `db` as the hostname) and I'm guessing `192.168.2.109` is the IP of your Docker Host; or you changed the docker-compose file to have `network_mode: host` on the `app` service but not the `db` service.

PR #151, opened 2019-05-25
CloudFlare DNS Challenge

OhHeyAlan:
Not sure how to route the certs to the backend of Nginx-Proxy-Manager but I can confirm that running this stack will allow LetsEncrypt/CertBot to do a DNS challenge via CloudFlare for wildcard certs.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-151-amd64`

jodfie:
sorry to be a bit of a dunce here... but after adding the LE docker and the conf and Env configured... how do i utilize this?

tomschlick:
Any status on this PR? This would be a great addition IMO.

jc21:
@jodfie makes a good point, I'd be happier to merge this if there was a new readme file in the `doc` folder specifically for Cloudfare DNS Challenge setup.

jc21:
Also, please change this PR to the `develop` branch.

jonfairbanks:
Hey @OhHeyAlan, can you make these changes? Would love to see this committed.

jodfie:
bumping this again to ask @OhHeyAlan if he can correct this so it can be implemented and also how to set it up?...

cristiangauma:
@jc21 is there any chance this will be merged or re-implemented soon?. DNS-Challenge would be awesome, as a lot of us don't publish our Proxy as it is part of our home/internal network.

cristiangauma:
I've been checking this PR. 

It only adds the necessary code to launch another container to do the DNS-Challenge and to download the certificates, but it doesn't add any _value_ to NPM itself using the app as the certificate manager using DNS Challenge. We already can download the certificates via multiple ways using externals tools (the container used in this PR, acme.sh, certbot installed locally...), which is what this PR is doing.

I've been also checking the NPM repo and if I could follow the code properly (I'm not a Node/Javascript person) I've not seen an automatic way to import .pem/.crt&.key from a directory. 

Maybe, if there would be a way to do it, this PR (or other tools) could work as we could share/use volumes to share the data between the certificates downloader and NPM.

Is there a way to import certificates from a directory?.

jc21:
No there isn't a way to do this - but a bash script that curl's the NPM API would be enough to add or update certs.

But yes I agree, adding another container to the stack doesn't really solve the root cause: the interface and flow doesn't support dns challenges well at all.

Issue #135, opened 2019-05-06
[NOOB] Prevent Direct Public IP access?


Kevinf63:
Hi there, I fully apologize for my lack of networking knowledge in advance.

I'm curious to know is there a method or implementation in NPM/nginx default configs that prevents **direct access** over my **public IP address** and instead only allows traffic routed through my Cloudflare **DDNS**?

Perhaps being able to globally alter the default config to deny all but whitelist these IP's?
[https://www.cloudflare.com/ips/](https://www.cloudflare.com/ips/)

I figured it might help my poor Virgin Media SuperHub 3.0 router falling over from port scanning/bots DDOSing, but I'm totally naive on this subject to be honest and could be a simple fix staring me in the face!

jc21:
Well sure you could use the advanced tab in proxy hosts to specify your whitelist options, but if you're using LetsEncrypt certificates with a whitelist, then it's likely not to work and you'll have to obtain a whitelist of LE servers as well.

Kevinf63:
Cheers for the quick response! No LetsEncrypt certs being used directly on NPM currently (HTTP only). Just the certs on the Cloudflare side although I'm not sure if thats unwise.

If I use a deny all; and allow those individual IP ranges it should work right? I seem to hit a snag were it seems the deny all is taking precedence over the other rules, likely an order or nginx syntax stumbling block.

jc21:
When I use something like this:
```
allow 203.174.180.157;
deny all;
```
it works as expected for me

the1ts:
I was using the same style as above in location /
`allow 192.168.0.0/24
deny all`
But this was also denying renewing of certs, I fixed by adding an allow all to the letsencrypt-acme-challenge.conf alongside the turning off auth:
`^~ /.well-known/acme-challenge/ {
    auth_basic off 
    allow all`
Am I doing something wrong or could this be a new default entry without harm?

jc21:
True I didn't consider the acme challenge, should allow all regardless of custom configuration. If you PR that it would be great. I've been too busy lately :/

PR #165, opened 2019-07-04
Open up lets Encrypt acme challenge config

the1ts:
As discussed in Issue #135. Since Lets Encrypt don't publish IP ranges that their acme challenge service will be sourced from, we need to allow free access to this location special to override any IP ACLs added by Advanced Custom Nginx Configuration. Due to the way Nginx config is applied, this only applies to the regex and below, keeping the IP ACLs working for the rest of the website.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-165-amd64`

Issue #233, opened 2019-11-09
Forward Hostnames longer than 50 characters are truncated


syntheticgoo:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
YES
- Are you sure you're not using someone else's docker image?
YES
- If having problems with Lets Encrypt, have you made absolutely sure your site is accessible from outside of your network?
N/A

**Describe the bug**
- A clear and concise description of what the bug is.
In the admin dashboard for proxy host, if you edit/add a host with a domain longer than 50 characters in the 'Forward Hostname / IP' field, it is truncated. If you edit the full hostname manually in the database and then attempt to re-save using the UI dashboard, a popup message appears with the message: "data.forward_host should NOT be longer than 50 characters".
- What version of Nginx Proxy Manager is reported on the login page?
v2.0.14

**To Reproduce**
1. Add a Proxy Host in the Dashbaord
2. Enter a long Hostname
3. Save Proxy Host
4. View the truncated hostname Destination in the Proxy Host list

**Expected behavior**
Longer proxy host domains should be allowed in the UI.

syntheticgoo:
The issue is this setting in src/backend/schema//endpoints/proxy-hosts.json

`    "forward_host": {
      "type": "string",
      "minLength": 1,
      "maxLength": 50
    },
`

capitangiaco:
what's the right way to change it on a running server ?

jc21:
It's baked in to the image, so you either have to wait for a fix or find the file in the docker image and create your own docker image that extends from mine, overwrite the file and run that. Might be easier to wait. I'll take a look today

jc21:
Actually this was fixed in 2.2.4. #mondays

capitangiaco:
I am on **2.3.0**, and in the file: /var/lib/docker/overlay2/26fae2ff7dcca2d17edf07ff67ae3590c909a408ed34915151543993d53058f5/merged/app/schema/endpoints/**proxy-hosts.json** line seems corrected: _"forward_host": { "type": "string", "minLength": 1, "maxLength": **255** },_
But I cannot insert url longer than 50 characters in Forward Hostname / IP*

jc21:
Is this for a proxy host or a redirection host or other? I was able to create more than 50 chars
[...]

PR #234, opened 2019-11-09
Fix hostname max length (#233)

syntheticgoo:


jc21:
As there are a lot of path changes in 2.1.0, this PR will need to be rebased manually.

PR #240, opened 2019-11-13
Increased nice_name field length to allow a larger number of LE certificates generation

EugeneKharasov:
We experienced certain issues while implementing nginx manager in our environment. Notably, since we are using microservices infrastructure, we run a large number of services requiring to use the same SSL certificate. Currently, nginx manager supports only 15 concurrent certificates generation via Letsencrypt, so we made it 50.


jc21:
Instead of editing the initial migration file I strongly recommend creating a new migration file that alters the table field. This way everyone who has already deployed NPM will get the change, not just the newcomers.

jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-240-amd64`

jc21:
If you're still considering doing this PR, please open again against the `develop` branch.

PR #256, opened 2019-12-15
[WIP] added experimental multidb support

### LINKED ISSUE #185: Updating the database client
Would it be possible to update the database client so we can use MySQL 8? Currently, it does not work due to unsupported authentication mode.

### LINKED ISSUE #200: No-config databases or json data storage
FWIW, it would be nice to remove the need for database configuration, or an extra database, by using redis, couchbase, mongodb or other nosql databases, or better yet just store/persist the data in a json file that can be stored on a volume or NFS share.  Configs will not change often enough to justify a full database with its added complexity and overhead.

tg44:
It's probably not a single-commit change, but I will summarize what I done;
  - I made the Dockerfile to work as a docker file (the build is stateless)
    - this should be either reverted or copied over to the other archtypes
    - also added a dockerignore
    - it was a "must" bcs I can't make my mac to build this project correctly
    - probably a "how to contribute" readme would solve this also
  - updated the knex dep
    - also went to mysql2
    - also added the sqlite3
    - also added the pg
  - added a `knex-native` engine (`db.js` and `install.md`)
  - fixed the migration (`20190227065017_settings.js`)
    - as I understand we start a db and migrate it, and while we in a transaction to migrate we also try to write the db from an another (or maybe from the same?) db connection. This is simply not working with sqlite bcs you have only one thread/connection/whatever to write the db. Also it seems to be a bad thing according the knox github issues.
  - moved the insert statement from the migration to the setup (`setup.js`)
    - also refactored the file a bit so its probably more readable
  - added a `now_helper`
    - bcs sqlite is a dummy
  - used the `now_helper` instead of the `Model.raw`

If you want atomic commits I think the correct steps would be:
  - write a developer doc OR fix docker
  - factor out now_helper
  - factor out + fix the migration and refactor the setup
  - add the new engine + deps + extend the now_helper

It is working with sqlite3 right now. I didn't test mysql, and pg. (Also probably would be nice to make the compose stack to support examples to all three and add default configs to all three.)

(I love the idea and the functionality BTW, really nice app!)

Probably fixes #185 and fixes #200


jc21:
If you're still considering doing this PR, please open again against the `develop` branch.

tg44:
Maybe I'm the sensitive, but this is kinda rude. I made a PR ~3 month ago. You don't even comment it, and now I should reopen it?

Ofc I will do it bcs I want to work with sqlite instead of mysql, but still, at least you could mention if its a good or a bad set of modifications...

jc21:
Honestly, I haven't had time to invest in this project over the last 6 months. I'm happy for this change to go in once it's fully tested, but since the CI was down at the time you submitted the PR, there hasn't been a docker image for anyone to test with.

jc21:
CI is back though, your next pr will be built for you and others to test.

LouisSung:
Hi, @jc21 and @tg44 
Thank you two for making this project and sending PR
I still have no time on trying this project, but I'm surveying a GUI-based solution for Nginx.
However, run an external database might cost performance, that's why I wonder if SQLite is supported and this PR is still available.
Or there's any reason to use a standalone database rather than SQLite? Since I don't know if there are any constrains or just the preferences?

Thank you~

tg44:
I think it can be recreated based on this PR. I run a mysql server right now (for basically no reason :( ). If you need further help, I try to help, just summon me to some comment section.

PR #207, opened 2019-10-08
Fix wrapping when too many hosts are shown

margaale:

PR #293, opened 2020-02-19
v2.1.0

### LINKED ISSUE #279: JWT Token expiration
I have a question about the access tokens issued by the api. It doesn't seem like they expire. Is this right? I am trying to integrate the manager into my own web app to seamlessly add hosts and generate SSLs but I can't figure out how to get the token to issue with an expiration from the REST endpoint "/token". The payload contains a secret and identity property and tried adding the "expiry" property which returns an "expires" property with the value I would expect but the token keeps functioning after that expires date has passed. 

I know this might me outside the scope of how you want the manager to be used but it would be a great help if someone could point me in the right direction. Thanks.

jc21:
Now using `docker buildx` for multiarch: amd64, armv7, arm64. No longer building for armv6. Let me know if you want more arch support in future.

- All branches and PR's will be multiarch builds
- Using a new [docker base image](https://github.com/jc21/docker-alpine-nginx-full)
  - nginx v 1.17.8 with ipv6 support
  - alpine 3.11.3
- JWT expiry changed to 1 day, fixes #279 
- Added Cypress Testing
- Refactored project structure to better organise frontend and backend
- Updated frontend dependencies
- Added backend code formatting rules and enforce them in CI
- Opened up [CI for public viewing](https://ci.nginxproxymanager.jc21.com) and added badge to readme
- Removed importing code for version 1 installations
- UI wrapping improvements (#207)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-293`

Issue #356, opened 2020-04-09
Restricting Access by IP Address


Indemnity83:
**The Problem**
I'm using the reverse proxy to give simple domain names to a couple of dozen services all running on docker in my home so I don't have to remember what random port the container's web interface is on. I suspect this use case is quite common (see #137, #135 as similar examples). However, this potentially exposes my proxy hosts to the internet at large, and I don't want public access to these services. 

The current Access list will let me put a basic auth "firewall" between the outside world and my private services, but it also complicates access for legitimate access (particularly since password managers can't always fill out basic auth requests, and strong passwords are highly encouraged for anything public). 

**The Solution**
Expand the Access Lists to allow restriction by IP address in addition to basic auth. The NGINX documentation actually [has an amazing example](https://docs.nginx.com/nginx/admin-guide/security-controls/configuring-http-basic-authentication/#combining-basic-authentication-with-access-restriction-by-ip-address) that describes exactly what we're after. 

I've made a mockup of the revised "New Access List" modal (the HTML can be found [in a gist here](https://gist.github.com/Indemnity83/12a6a0f34127b79b8200ff43982fe8ce)).

![ezgif com-optimize](https://user-images.githubusercontent.com/35218/78948259-8e4a7880-7a7c-11ea-973a-d5d94dc10726.gif)

 The goal of the UI changes is to keep things simple and approachable even for users who may not understand what's going on under the hood.  No other UI changes would be necessary. I'm not certain of the extent of back-end changes required at this point. 

**Alternatives**
This configuration can be achieved by utilizing the advanced -> Custom Nginx Configuration feature already built into the application. However, there are a few issues/concerns with implementing IP restriction this way:

1. It requires the user to have an understanding of NGINX configuration
2. Changes/updates must be applied to every proxy host individually 
3. It's not obvious from the proxy hosts list that there is any access control enabled (or conversely if you've forgotten to apply access restrictions to a host)

Indemnity83:
I'm willing to work on a PR on this as well, but I'm not intimately familiar with the js libraries being used so I may only be able to provide an 80% solution before needing some assistance.

Thijmen:
Awesome @Indemnity83 ,  I was looking for this as well. Is there any way I can help you with? I am not familiar either, but I am sure we can figure it out!

Indemnity83:
This was merged in with #360  🎉

Yabbo:
[...]

PR #360, opened 2020-04-11
Client Access Lists

Indemnity83:
This PR adds client (IP address) based access control to the application in a way that aims to cover the 80% case based on #356 

There are a few significant outstanding tasks remaining to make this ready for showtime. 

- [x] Address validation needs to be expanded to support IPv4, IPv6, CIDR notation and the 'all' keyword
- [x] The backend needs to validate that either an Authorization or Access rule has been provided
- [x] The Nginx config needs to actually reflect the access list that is defined
- [ ] ~~Write tests~~

Screen Shot 2020-04-11 at 12 34 48 AM
Screen Shot 2020-04-11 at 12 34 59 AM
Screen Shot 2020-04-11 at 12 35 15 AM
Screen Shot 2020-04-11 at 12 35 45 AM
Screen Shot 2020-04-11 at 12 36 07 AM


jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-360`

jc21:
Thanks for doing the hard work here, greatly appreciated. I'm using your code and I get the following error on save:

Screen Shot 2020-04-14 at 8 49 09 am

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-360`

Indemnity83:
@jc21 happy to do it, I would really benefit from the feature. I intentionally marked this PR as WIP because there are a couple of fairly significant items that need to be finished to make this complete. You've hit on one of them. 

I'm not sure what the best way to handle that one is either. It doesn't look like the JSON API spec supports multiple validation formats? I could do it as one big regex, but that would get really ugly really fast to make a single regex that could validate Ipv4, ipv6 and the CIDR notations of both.

I didn't see an example in the codebase of doing validation in what I'll call the controller either? I may just have not known where to look. If you can provide an example of doing validation outside of the JSON schema I'll be happy to implement it.

jc21:
So this line here is where the JSON payload is validated in the API Handler:

https://github.com/Indemnity83/nginx-proxy-manager/blob/ip-access-control/backend/routes/api/nginx/access_lists.js#L59

And this is the json schema validation for that one as you already know:

https://github.com/Indemnity83/nginx-proxy-manager/blob/ip-access-control/backend/schema/endpoints/access-lists.json#L72

for the `address` field, yes you can have multiple validations like so:

```json
    "address": {
      "oneOf": [
        {
          "type": "string",
          "format": "ipv4"
        },
        {
          "type": "string",
          "pattern": "^regexHere$"
        },
        {
          "type": "string",
          "pattern": "^anotherRegexHere$"
        }
      ]
    },
```

Indemnity83:
Perfect, the `oneOf` directive was what I didn't know existed :)

Is there something similar to make sure that either an "item" or "client" is provided? I have them both set as `"minItems": 0` right now. The frontend checks that at least one is provided before submitting, but it should be checked in the backend too.

jc21:
AJV is the validation library I'm using and it has some great simple documentation for your options: https://github.com/epoberezkin/ajv#validation-keywords

Yes you can make sure that either one of them is completed in the json schema, but the only way to do it is really stupid and you have to wrap what you have in a "oneOf" array and duplicate a lot of the json schema as is. So I don't recommend that. Instead I would just have a quick check here:

https://github.com/Indemnity83/nginx-proxy-manager/blob/ip-access-control/backend/internal/access-list.js#L25

to make sure that the length of the items in the `data` are > 0 in one of those places.

jc21:
ie:

```javascript
	create: (access, data) => {
		return access.can('access_lists:create', data)
			.then((/*access_data*/) => {
				if ((typeof data.items === 'undefined' || !data.items.length) && (typeof data.clients === 'undefined' || !data.clients.length)) {
					throw new error.InternalValidationError('At leaste one user/pass or address must be defined');
				}
				
				// ...
			}

```

jc21:
Docker Image for build 4 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-360`

jc21:
Docker Image for build 5 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-360`

Indemnity83:
This should be fully functional now.

It could use a pretty though peer review as I'm sure there is some refactoring that could take place. in particular, this expansion feels gross but I don't see a cleaner solution. 

https://github.com/jc21/nginx-proxy-manager/blob/f5ee91aeb3756b824aea32db8bfe72f8d4bb8604/backend/internal/access-list.js#L72-L75

jc21:
Awesome. Just verified it works as expected! I'll be able to fix up any "mess" later. Thanks heaps for the contribution!

Thijmen:
@Indemnity83 would it also be possible with this feature to allow the IP's in the list and if the IP is not in the list, provide the HTTP Auth?

Indemnity83:
That’s exactly what the “satisfy any” option does.

Thijmen:
Totally missed that, thanks! Great addition to the repo, awesome work man!

Issue #357, opened 2020-04-10
Adding AWS S3 bucket hosting as an option


dpanesso:
**Is your feature request related to a problem? Please describe.**
We manage around 30 hosts on our nginx reverse proxy. About 1/3 of those are hosted on AWS S3. We need https on almost all of them and using something like Cloudfront to have SSL would be overkill.

**Describe the solution you'd like**
Add an option on the Proxy hosts to use S3 as a forward option. This would change the location block to use the options needed for it to work. Something like this is what we use:

```
        proxy_http_version     1.1;
        proxy_set_header       Connection "";
        proxy_set_header       Authorization '';
        proxy_set_header       Host bucket.s3-website-us-east-1.amazonaws.com;
        proxy_hide_header      x-amz-id-2;
        proxy_hide_header      x-amz-request-id;
        proxy_hide_header      x-amz-meta-server-side-encryption;
        proxy_hide_header      x-amz-server-side-encryption;
        proxy_hide_header      Set-Cookie;
        proxy_ignore_headers   Set-Cookie;
        proxy_intercept_errors on;
        add_header             Cache-Control max-age=31536000;
        proxy_pass             http://bucket.s3-website-us-east-1.amazonaws.com/;
```

The maximum length of the forward_host variable would also have to be increased as 50 is too small.

**Describe alternatives you've considered**
I considered using the custom location option but, the host parameter is set in the non editable section of proxy.conf  

**Additional context**
This would be my proposal:

![image](https://user-images.githubusercontent.com/2687121/78966237-dee5c400-7ac4-11ea-8f9f-e512f7715bd4.png)

I have made some progress on development. I can open a PR to check out my progress. I would consider some additional validations for it to be complete.

PR #367, opened 2020-04-14
Adding S3 host option

dpanesso:
This PR allows to define an S3 host. We use this very often to serve static hosting from our AWS S3 buckets. It references #357.

It uses this location block to serve the S3 bucket content:

```
        proxy_http_version     1.1;
        proxy_set_header       Connection "";
        proxy_set_header       Authorization '';
        proxy_set_header       Host bucket.s3-website-us-east-1.amazonaws.com;
        proxy_hide_header      x-amz-id-2;
        proxy_hide_header      x-amz-request-id;
        proxy_hide_header      x-amz-meta-server-side-encryption;
        proxy_hide_header      x-amz-server-side-encryption;
        proxy_hide_header      Set-Cookie;
        proxy_ignore_headers   Set-Cookie;
        proxy_intercept_errors on;
        add_header             Cache-Control max-age=31536000;
        proxy_pass             http://bucket.s3-website-us-east-1.amazonaws.com/;
```

The maximum length of the forward_host variable would also have to be increased as 50 is too small.

This is what it looks like:

image

### DIFF
--- a/backend/migrations/20200409182508_s3_host.js
+++ b/backend/migrations/20200409182508_s3_host.js
@@ -1,0 +1,42 @@
+const migrate_name = 's3_host';
+const logger       = require('../logger').migrate;
+
+/**
+ * Migrate
+ *
+ * @see http://knexjs.org/#Schema
+ *
+ * @param {Object} knex
+ * @param {Promise} Promise
+ * @returns {Promise}
+ */
+exports.up = function (knex/*, Promise*/) {
+
+	logger.info('[' + migrate_name + '] Migrating Up...');
+
+	return knex.schema.table('proxy_host', function (proxy_host) {
+		proxy_host.integer('s3_host').notNull().unsigned().defaultTo(0);
+	})
+		.then(() => {
+			logger.info('[' + migrate_name + '] proxy_host Table altered');
+
+			return knex.schema.table('dead_host', function (dead_host) {
+				dead_host.integer('s3_host').notNull().unsigned().defaultTo(0);
+			});
+		})
+		.then(() => {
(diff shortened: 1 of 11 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-367`

jc21:
Awesome stuff. Can I suggest that instead of having a switch to say "Is this an s3 host", we could simply detect that the `forward_host` is an s3 hostname and make the nginx inclusions at that point. This reduces the risk that someone enables the option for non-s3 hosts and vice versa.

We could just check for something like `.*s3.*amazonaws\.com$`

dpanesso:
That's a very interesting idea! Would be much better to auto detect for sure. 

I had some time today and did dome tests, and now I am confused, because during my tests I had an error and my location block was wrong but it still worked. So, it seems AWS has changed something and doesn't require the headers or the host to be the bucket generated URL. I seems this PR won't be actually necessary. :neutral_face:

Let me confirm with some more tests and will either close the PR o adjust with your recommendations.

apignard:
autodetection will only work if S3 is AWS.
If you want autodetect add a way to add regex or such for add custom domain for S3 storage.
(other public cloud, on premise, private cloud etc have also S3 storage)

jc21:
Closing due to inactivity, and probably isn't required.

Rustymage:
How would one go about a Minio (S3 Complaint) instance redirection to the correct URL?

PR #373, opened 2020-04-16
Improve terminal script colors

Indemnity83:
If a terminal cannot interpret \e[* color codes, then it will be "polluted" which makes output harder to read. tput will handle escape characters depending on the terminal capabilities.

Screen Shot 2020-04-15 at 10 04 19 PM
Screen Shot 2020-04-15 at 10 03 51 PM

### DIFF
--- a/scripts/buildx
+++ b/scripts/buildx
@@ -1,12 +1,8 @@
 #!/bin/bash
 
-CYAN='\E[1;36m'
-YELLOW='\E[1;33m'
-BLUE='\E[1;34m'
-GREEN='\E[1;32m'
-RESET='\E[0m'
+. $(dirname "$0")/log
 
-echo -e "${BLUE}❯ ${CYAN}Building docker multiarch: ${YELLOW}${*}${RESET}"
+log "Building docker multiarch: ${YELLOW}${*}" "info"
 
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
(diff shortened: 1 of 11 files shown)


jc21:
Yeah I used to use tput but not all docker containers had it installed for my scripts. In this case, for this project, we can go with it.

Indemnity83:
Let me fix my git history not sure what happened there. 

Also looks like as you mentioned the docker containers don't have tput

jc21:
The build is failing because TERM env var is not set in the docker build. Just add
```
ENV TERM=xterm-256color
```
to line 17 of the [Dockerfile](https://github.com/Indemnity83/nginx-proxy-manager/blob/script-colors/docker/Dockerfile) and the build should be ok

jc21:
This is failing CI because of unrelated reasons I introduced recently. A rebase will fix it after I fix the branch

Indemnity83:
I probably should have just squashed the commits at this point, but I'll just let the build finish to make sure it's working properly and you can decide to squash and merge or just merge.

jc21:
You'll have to merge `jc21:develop` to your branch and push again for the build to work I think

Indemnity83:
went a different route; extracted a log helper to make changes more universal but I also found that using `printf` instead of `echo -e` seems to work in more places and shouldn't have the problem that tput had with dockers lack of a terminal.

jc21:
Closing due to inactivity

PR #376, opened 2020-04-16
expand the maximum size of the forward_host

spalger:
I'm trying to setup a proxy host to a hostname that is 56 characters long and that's not allowed by the schema for the proxy host for some reason :/ I think it's probably best to avoid arbitrary limits like this so I've removed it from the schema, but I could also set it to 255 or something if you prefer.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-376`

jc21:
I'd prefer the limit to match the database:

`forward_host varchar(255) NOT NULL`

otherwise when someone exceeds the 255 db limit (and someone will) depending on the database version the value will either be clipped without warning or they'll see a sql error.

jc21:
Thanks, I'll merge once CI passes.

jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-376`

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-376`

jc21:
This will be in the next minor release but until that comes along you can use this docker image:
`jc21/nginx-proxy-manager:github-develop`

PR #402, opened 2020-05-09
Fix address validation rule to allow 'all' keyword

Indemnity83:
The rule was looking for the keyword 'any' but should have been looking for 'all' 

http://nginx.org/en/docs/http/ngx_http_access_module.html


jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-402`

PR #403, opened 2020-05-09
Don't ask for username/password if none are defined

Indemnity83:
This commit optimizes the Nginx config implementation of an access list, preventing an auth-check if no users are actually defined.

This should still mean that creating an empty access rule results in an inaccessible site because of the default `deny all` rule (secure by default). But prevents Nginx from even processing Basic Auth, or asking for a username and password if there are no users to check against. 

~This PR also fixes a bug that was preventing Auth rules from being applied to a site after enabling a previously disabled site.~ (moved to #407)

### DIFF
--- a/backend/internal/access-list.js
+++ b/backend/internal/access-list.js
@@ -71,7 +71,7 @@ const internalAccessList = {
 				// re-fetch with expansions
 				return internalAccessList.get(access, {
 					id:     data.id,
-					expand: ['owner', 'items', 'clients', 'proxy_hosts.access_list.clients']
+					expand: ['owner', 'items', 'clients', 'proxy_hosts.access_list.clients', 'proxy_hosts.access_list.items']
 				}, true /* <- skip masking */);
 			})
 			.then((row) => {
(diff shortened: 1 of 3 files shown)


andyjhall:
If you're only passing in the item list in order to ensure its greater than zero would it make more sense to just pass a bool in instead?

Indemnity83:
There are actually a couple places where I end up counting the number of items, I can pull that common logic into the access-list model easily enough. Since the template is the only place I end up checking if it's greater than 0 I don't think it makes sense to pull that up.

Indemnity83:
Actually, now that I look at it there is only one use in the backend, and one use on the front-end. Pulling the counting logic into the back-end model and presenting that data to the front-end would require changes to the API to add information that is easily inferred and I don't think it makes any significant improvement to the readability or maintainability of the code.

jc21:
These changes seem to conflict with #407 ?

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-403`

Nedlinin:
👏 Looking forward to this making it in.  Appears to fix #419

andyjhall:
Looks to be working here 👍🏻

Indemnity83:
> These changes seem to conflict with #407 ?

Not intentionally; I tried to pull out the conflicting bits into #407. I'm happy to rebase and de-conflict either PR after one of them is merged. I wanted to keep the PRs self-contained.

geertmeersman:
Hi, I just ran the last version (docker v2.3.0) and now the access rules (IP list) is not included anymore... Am I missing something?

Subv:
@geertmeersman Hi! Could you please try #435?

geertmeersman:
> @geertmeersman Hi! Could you please try #435?

Hi @Subv , I just tested the docker image jc21/nginx-proxy-manager:github-pr-435 and confirm it is working as expected! Thx

PR #405, opened 2020-05-10
fix spelling of 'satisfy'

Indemnity83:
I think this is the last remaining place of the misspelling.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-405`

PR #435, opened 2020-05-30
Don't use duplicate relations when eager-loading access list items and clients.

### LINKED ISSUE #434: 2.3.0 - Access list allow rules not applying to nginx config files.
**Checklist**
Yes
Yes
- If having problems with Lets Encrypt, have you made absolutely sure your site is accessible from outside of your network?
Yes

**Describe the bug**
2.3.0 does not apply the ALLOWED IP access lists to the individual nginx config files.

**To Reproduce**
1) Create access list with IP address with ALLOW rule.
2) Attach rule to proxy_host
3) Observe that no Allow rule is created above the deny all rule.

**Expected behavior**
Specified allow rules should apply above the default deny all; 

Issue not present in 2.2.4 branch.

Subv:
This fixes an Objection warning: `Duplicate relation "proxy_hosts" in a relation expression. You should use "a.[b, c]" instead of "[a.b, a.c]". This will cause an error in objection 2.0`.

It also fixes the access list clients not being properly eager-loaded when building the proxy host nginx configuration files, resulting in no IP access lists being printed into the file.

Closes #434


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-435`

jc21:
Nice Thanks :)

geertmeersman:
Hi, I just tested the docker image jc21/nginx-proxy-manager:github-pr-435 and confirm it is working as expected! Thx

PR #464, opened 2020-06-18
Update location-item.ejs - forward_host size increase to 200

vrenjith:


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-464`

PR #510, opened 2020-07-19
Multidb - sqlite support

### LINKED ISSUE #185: Updating the database client

### LINKED ISSUE #200: No-config databases or json data storage

tg44:
I recreated this PR; https://github.com/jc21/nginx-proxy-manager/pull/256

Separated commits;
 - The first solves the ~~transaction in transaction~~ connection in connection problem in the migration, which causes problems on db-engines which has no multithread support (like sqlite3), also adds a minor refactor to the setup.js
 - The second adds a helper to the 'NOW()' sql function, bcs some db engines (like sqlite) has another syntax for this...
 - The third adds a new db-config option (`knex-native`) and let an experienced knex user to add some own knex-native configurations. Also, this commit showcases an sqlite3 config, and adds psql and sqlite3 libs.

Probably fixes #185 and fixes #200

Missing things;
 - document the sqlite config
 - try out the psql support

### DIFF
--- a/backend/setup.js
+++ b/backend/setup.js
@@ -5,111 +5,145 @@ const logger              = require('./logger').setup;
 const userModel           = require('./models/user');
 const userPermissionModel = require('./models/user_permission');
 const authModel           = require('./models/auth');
+const settingModel        = require('./models/setting');
 const debug_mode          = process.env.NODE_ENV !== 'production' || !!process.env.DEBUG;
+
+function setupJwt(resolve, reject) {
+	// Now go and check if the jwt gpg keys have been created and if not, create them
+	if (!config.has('jwt') || !config.has('jwt.key') || !config.has('jwt.pub')) {
+		logger.info('Creating a new JWT key pair...');
+
+		// jwt keys are not configured properly
+		const filename  = config.util.getEnv('NODE_CONFIG_DIR') + '/' + (config.util.getEnv('NODE_ENV') || 'default') + '.json';
+		let config_data = {};
+
+		try {
+			config_data = require(filename);
+		} catch (err) {
+			// do nothing
+			if (debug_mode) {
+				logger.debug(filename + ' config file could not be required');
+			}
+		}
+
+		// Now create the keys and save them in the config.
+		let key = new NodeRSA({b: 2048});
+		key.generateKeyPair();
+
+		config_data.jwt = {
+			key: key.exportKey('private').toString(),
+			pub: key.exportKey('public').toString()
+		};
+
+		// Write config
+		fs.writeFile(filename, JSON.stringify(config_data, null, 2), (err) => {
+			if (err) {
+				logger.error('Could not write JWT key pair to config file: ' + filename);
+				reject(err);
+			} else {
+				logger.info('Wrote JWT key pair to config file: ' + filename);
+
+				logger.warn('Restarting interface to apply new configuration');
+				process.exit(0);
+			}
+		});
+
+	} else {
+		// JWT key pair exists
+		if (debug_mode) {
+			logger.debug('JWT Keypair already exists');
+		}
+
+		resolve();
+	}
+}
+
+function setupDefaultUser() {
+	(userModel
+		.query()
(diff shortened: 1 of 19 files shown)


tg44:
I think it is a CI error and not an error with the code...

backend/setup.js
```diff
@@ -5,111 +5,145 @@ const logger              = require('./logger').setup;
...
-				logger.debug('Admin user setup not required');
+			} if (debug_mode) {
+				logger.debug('Default setting setup not required');
 			}
 		});
+}
+
+module.exports = function () {
+	return new Promise((resolve, reject) => {
+		return setupJwt(resolve, reject);
```
jc21, line 143: setupJwt is not a promise, so returning it as is will be async but won't reject on error properly. If it were a promise, you wouldn't need to wrap it in a Promise constructor either.

backend/setup.js
```diff
@@ -5,111 +5,145 @@ const logger              = require('./logger').setup;
...
+					logger.info('Initial admin setup completed');
+				});
+		} else if (debug_mode) {
+			logger.debug('Admin user setup not required');
+		}
+	});
+}
 
-				return userModel
+function setupDefaultSettings() {
```
jc21, line 117: Not quite sure why you took this from a Migration and put it here, you're not doing that to any other migrations...

jc21:
Docker Image for build 4 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-510`

backend/setup.js
tg44: This is the only migration where the concept was broken. I didn't needed to do it in any other migration bcs all of the other migrations just modify the schemas, and not trying to insert things to it. (BTW; we insert users in the setup but we insert settings in a migration?)

As I understand when you run the migrations it connects to the db. And when you run the `settingModel.query().insert()`  it tries to connect to the db again (while we still migrate). Which in multithreaded env (like pg or mysql) not a big deal, you can have multiple connections/connection pools to the same db. But in sqlite you hava a lock on the file, so while one connection is attached to the db, the other connection "can't" connect to it.

If you are interested in it deeper I referenced an issue in the commit message.

backend/setup.js
tg44: I literally just extracted the code to a new function. I think if it has some logical error now, it had it in the past too.

Probably I could have moved the promise creation to the `setupJwt` too, but probably it is just a matter of taste.

tg44: ohhh, or I just simply not need to write the "return" keyword in this case?

tg44:
Starting from today I use this image on my home server.

backend/setup.js
jc21: Not really, it just could be written differently but let's just get it through for now

backend/setup.js
jc21: Ah ok I get your point about breaking the concept. All good.

To clarify, "migrations" should be changes that occur before the App really runs. It's used on the Knex layer and should not perform any logic on data and assumes the database is in a certain state prior to execution.

Setup users is part of the app because it has to check if there is an admin user and only when not found, it creates one. As this happens on startup each time, a migration is not the right place for that.

tg44: I understand this, BUT if injecting users is a runtime check, inserting settings why not? Also if inserting settings is essential to start the app, I think the default user is essential to start the app too. It breaks the logic if one or the other handled differently. I think it's mostly a theoretical question if inserting default data is part of the db migration or not, and most of the JVM world would place them to the migrations (including myself), but here the code already introduced both ways, and I simply chose the comfy side.

tg44:
Thanks the merge!

PR #544, opened 2020-08-06
Fixed now_helper for sqlite (time is missing)

jlesage:
The sqlite `date('now')` function returns only the current time.
```
# echo "SELECT date('now')"|sqlite3
2020-08-06
```
To be comparable to the `NOW()` function of mysql, the date and time should be returned.  The `datetime('now','localtime')` function of sqlite does this:
```
# echo "SELECT datetime('now','localtime')"|sqlite3
2020-08-06 16:44:59
```
Which is the same thing as mysql:
```
# echo "SELECT NOW()"| mysql
NOW()
2020-08-06 16:44:59
```

### DIFF
--- a/backend/models/now_helper.js
+++ b/backend/models/now_helper.js
@@ -6,7 +6,7 @@ Model.knex(db);
 
 module.exports = function () {
 	if (config.database.knex && config.database.knex.client === 'sqlite3') {
-		return Model.raw('date(\'now\')');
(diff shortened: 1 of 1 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-544`

PR #558, opened 2020-08-14
Static site hosting and location functionality

lopugit:
I made some modifications to the proxy hosts flow which adds a static toggle which switches between proxying and static root/alias/try_files based locations. 

I've modified the base proxy_hosts form and the locations one so you can mix and match between proxy hosts and locations and static hosts and locations. 

There's a little bug because I'm not so experienced with Backbone marionette and I used the static toggle to change the required fields on the form from forward_host and forward_ip to index_file and root_dir. And to rerender the form to show that update I ran a this.render call which seems to break the location items rendering so you have to save the root config for a proxy host before clicking edit again to do the locations.

Here's an album of visual changes

https://imgur.com/a/qgxzYFu

You can see the template changes in `_location.conf` `_listen.conf` and `proxy_host.conf` 

I had some trouble with the booleans either being 0 or 1, because in the listen forms the model values are used but in the proxy_host form the actual HTML form values are used which is integer on the checkbox's.

I think this pull request will open up nginx proxy manager to a lot more people, as SSL static hosting is one of the most sought after things on the web.

Needs a bit of polishing but that's why I pulled it into the develop branch.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-558`

jc21:
Thanks for doing all this work! I agree backbone and marionette is dated and won't be my choice for version 3.

Where this PR falls short is that a Proxy host should not deliver static content. It would be better to separate this static stuff into a new Host type. Hopefully this would avoid confusion and would definitely make validation rules nicer.

If you want to, I can create a new branch in this main project for this static feature, merge this PR to it and build on it from there?

lopugit:
> Thanks for doing all this work! I agree backbone and marionette is dated and won't be my choice for version 3.
> 
> Where this PR falls short is that a Proxy host should not deliver static content. It would be better to separate this static stuff into a new Host type. Hopefully this would avoid confusion and would definitely make validation rules nicer.
> 
> If you want to, I can create a new branch in this main project for this static feature, merge this PR to it and build on it from there?

True true.

I don't know how much time I'll have to do more work on it, I opted to just do my nginx stuff manually since it was only for a couple of simple projects. 

Yes a frontend framework like quasar/vue or react would help I think, the backend is bloody solid though good job on it and easy to develop and debug :) 

If I get time I'll split the static logic into a separate host, I was just being lazy to be honest haha. But the foundation is now there :) And I know what libraries are used for templating and etc.. so can contribute quicker now :) 

Sure if you make the branch for static that'd be awesome :)

jipjan:
I'm working on this in my own fork to make this a separate option as suggested by @jc21

PR #559, opened 2020-08-14
Removed the hardcoded `--webroot` certbot argument to better support DNS challenge

jlesage:
Also, note that this option is already set in the default `letsencrypt.ini`.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-559`

jc21:
Tested and seems ok. Certbot seems to be taking ages to request or renew certs but this appears to be the same prior to this change as well.

PR #587, opened 2020-08-30
Allow inputs to update

duhruh:
So there's this annoying bug in the modal for Add Custom Certificate where when you select a file to upload the input text doesn't reflect the name of the file that was chosen, making the user think nothing happened at all. However in investigating this I noticed that everything works fine with the form so this is just a visual feedback change.

cert


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-587`

Issue #663, opened 2020-10-16
Not possible to import custom certificate key using elliptic curve instead of rsa


mkochenough:
**Checklist**

- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?

Yes

- Are you sure you're not using someone else's docker image?

Yes

- If having problems with Lets Encrypt, have you made absolutely sure your site is accessible from outside of your network?

I'm not using Lets Encrypt. This is for network internal usage.

**Describe the bug**

I used Nginx Proxy Manager v2.6.1.

I tried to add a custom certificate. I gave it a name and added its key and the certificate itself. After clicking "Save" the an error message is displayed:

Upload failed: Certificate Key is not valid (Command failed: openssl rsa -in /tmp/9c58b1f7-db1a-4806-bc91-fb6ee547fb32/tmp -check -noout 140620246457672:error:0607907F:digital envelope routines:EVP_PKEY_get0_RSA:expecting an rsa key:crypto/evp/p_lib.c:469: )

This is understandable is the key uses an elliptic curve instead of rsa. The problem is in backend/internal/certificate.js in the checkPrivateKey method. There an external command is executed with "openssl rsa -in ....".

Executing the same command locally on the key shows the same error. Exchanging "rsa" with "ec" solves the problem locally.

Somehow NPM should detect if an RSA or an EC key is used an use the proper command. Or just check EC if RSA failed before.

MarceloLagos:
The check is easy, but for some reason the validation of the result is empty (when I test) if ec is used in the openssl check.
`let key_type = private_key.includes('-BEGIN RSA') ? 'rsa' : 'ec';`
`return utils.exec('openssl '+key_type+'  -in ' + filepath + ' -check -noout')`
`.then((result) => {`

The RSA check output is one line:
`RSA key ok`
while the EC check output is 2 lines:
`read EC key`
`EC Key valid.`

Whenever I check `result` when using ec, it's empty.
Any ideas?

MarceloLagos:
The problem was openssl command was throwing the EC validation output as error (even though is successful).
Adding 2>&1 at the end of the command will put error output in standard output and therefore the output will be captured either way. Also, the following string validation needs an additional condition since the message for EC keys doesn't contain "key ok".
I've created Pull #666 with these changes.

mkochenough:
@MarceloLagos You are right. Tested your branch and it works perfectly. Closing my pull request.

PR #665, opened 2020-10-17
Check also certificate keys using elliptic curve algorithms

mkochenough:
This should help with issue #663.


jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-665`

mkochenough:
Thanks for this hint. I used this docker image. It still fails to import the key. Unfortunately the error message is empty. I will dig a bit around to fix this.

MarceloLagos:
The error message is empty because it doesn't fail the openssl command. Instead it works, but the output is empty, therefore it fails the check to see if "key ok" is in there.
Furthermore even if that doesn't fail, the check will since the verification for EC keys doesn't return "key ok" anywhere. See my post on #663

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-665`

PR #666, opened 2020-10-18
Check key for RSA header otherwise use EC, and output fix.

MarceloLagos:
Performing a check on the uploaded key looking for `-----BEGIN RSA` (should it be less specific like just `BEGIN RSA`?).
Then modify the openssl command to:
1. Use either `rsa` or `ec` to validate the key.
2. Sending standard error to standard output to catch all output (ec validation seems to send it to output error even when successful)

Finally modify the result verification by adding the string to be found on EC key validation (`key valid`) which is different than the one in RSA validation (`key ok`).


jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-666`

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-666`

PR #687, opened 2020-11-03
Allow additional dns challenge dependencies

chaptergy:
This PR will allow dns challenge plugins to have additional dependencies which will also be installed. This should fix https://github.com/jc21/nginx-proxy-manager/issues/680. Furthermore it is made sure, that the credentials folder exists before writing to it to prevent the following error:

```
/bin/sh: can't create /etc/letsencrypt/credentials/credentials-1: nonexistent directory
```


jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-687`

jc21:
Docker Image for build 4 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-687`

PR #695, opened 2020-11-06
Adds autoremove of failed certificate creations in DB

chaptergy:
This PR prevents certificates from showing up in the web interface when they failed to be created. No audit log is written, since nothing was changed. Up until now, an audit log entry was written, when the user had to manually delete the certificate, but when it was 'created', no log was written.

Fix https://github.com/jc21/nginx-proxy-manager/issues/629.

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -216,6 +216,13 @@ const internalCertificate = {
 											return saved_row;
 										});
 								});
+						}).catch(async (error) => {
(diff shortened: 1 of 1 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-695`

PR #745, opened 2020-11-28
Allow nginx to resolve docker hostsnames

### LINKED ISSUE #203: Can't communicate by name on docker network
I currently have this setup in a user defined docker network, "dockerwp".  In that network I have several other docker containers running as backend.  I have setup npm and have created several proxy hosts to point to the backened containers, again in that dockerwp network.  If I set npm to talk to them by name (ie http://website1) it can't communicate.  If I use the same settings and put use the internal ip address of that container (ie 172.21.0.6) everything communicates as expected.  Since I'm spinning up and removing containers quite regularly I'm, wasting a lot of time looking up the IP, it would be nice to be able to communicate by name, which should be possible on a user defined network.

Mechazawa:
image

I've been unable to connect to docker hosts using their hostname even if they are on the same network. This fixes it by overriding the incorrect auto generated default value for the dns resolver. 

See: http://docs.docker.oeynet.com/engine/userguide/networking/configure-dns/
> Note: The DNS server is always at 127.0.0.11.

Fixes #203

### DIFF
--- a/docker/rootfs/etc/nginx/nginx.conf
+++ b/docker/rootfs/etc/nginx/nginx.conf
@@ -49,8 +49,8 @@ http {
 
 	access_log /data/logs/default.log proxy;
 
-	# Dynamically generated resolvers file
-	include /etc/nginx/conf.d/include/resolvers.conf;
+	# Used for resolving docker hosts
+	resolver 127.0.0.11;
 
 	# Default upstream scheme
 	map $host $forward_scheme {


iiLaurens:
+1 for this. This could be quite helpful.

jc21:
This isn't going to work for everyone. There are other considerations here, such as people using the image with orchestration software.

Docker is not meant to allow communicating to other services on the same host using the default network.

I personally run my docker containers in a defined docker network and I use the docker-compose defined service names to connect to those services. It's really simple to do and provides another level of security.

```bash
docker network create sesemestreet
```

```yml
version: '3'
services:
  npm:
    image: jc21/nginx-proxy-manager:2
    container_name: npm
    # ...

networks:
  default:
    external:
      name: sesemestreet
```

```yml
version: '3'
services:
  nzbget:
    container_name: nzbget
    image: linuxserver/nzbget
    # ...

networks:
  default:
    external:
      name: sesemestreet
```

Then using the hostname `nzbget` in NPM works as expected.

TarikHaci:
I do use the exactly same of @jc21 but with the new way of noting:

```
.....
    networks:
      - bridgewithnginx
      
networks:
  bridgewithnginx:
      name: nginxnetwork
      external: true
```

But it only worked after in Dockerfile I made Expose of the Port, for some reason the mapping on docker-compose doesn't work with nginxproxymanager it only works with external IP. I managed finally to add containername:ip to nginxproxymanager by using EXPOSE portnumber and inputting that portnumber into nginxproxymanager.

PR #773, opened 2020-12-12
Allow every type of privkey

MichelBaie:
It cause problems with self generated clouflare keys


jc21:
Since #774 is merged now, if you can't replicate your problem with that fix, please close this PR.

MichelBaie:
Thanks !

PR #774, opened 2020-12-14
Better custom certificate handling

chaptergy:
This should fix https://github.com/jc21/nginx-proxy-manager/issues/594 and fix https://github.com/jc21/nginx-proxy-manager/issues/759.
Private key types are now no longer explicitly checked within proxy manager but instead it is outsourced to openssl. Furthermore a warning notifying the user, that passphrase protected keys are not supported was added, as well as an error thrown by the backend when the check timed out (which is usually because the command waits for the passphrase input).

This PR is in direct conflict with PR https://github.com/jc21/nginx-proxy-manager/pull/773

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -615,18 +615,26 @@ const internalCertificate = {
 	checkPrivateKey: (private_key) => {
 		return tempWrite(private_key, '/tmp')
 			.then((filepath) => {
-				let key_type = private_key.includes('-----BEGIN RSA') ? 'rsa' : 'ec';
-				return utils.exec('openssl ' + key_type + ' -in ' + filepath + ' -check -noout 2>&1 ')
-					.then((result) => {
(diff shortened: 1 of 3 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-774`

chrismcneil:
Thankyou, I've been banging my head against the wall all day, came to check for issues and saw this pr. Was able to successfully import my key :)

jc21:
Thanks for this fix. I've been on holiday and this was on my todo list. I think we should add some automated testing for this too.

webnex:
> Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-774`

I'm a complete noob :-) can you please advise me exactly what I need to do and how to set it up.
Many thanks to Michal

PR #842, opened 2021-01-25
Use configured default page also for IPv6

Shuro:
Just a small check for the ipv6 variable, similar to _listen.conf,
so that the configured default page is also delivered on ipv6 requests.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-842`

PR #849, opened 2021-01-28
Move custom http.conf include clause above include clauses for NPM-generated files.

MooBaloo:
Moved the include clause for the custom http.conf file above the include clauses for the NPM-generated files. Allows for more flexibility with adding custom nginx .conf files in /data/nginx/custom/.

Use case: adding configuration to the http block that is later referenced by other custom nginx .conf files in /data/nginx/custom/.

Example: adding a new log_format in /data/nginx/custom/http.conf, and then referencing it in an access_log clause in /data/nginx/custom/server_proxy.conf, which is included in the NPM-generated .conf files under /data/nginx/proxy_host/. 

Cannot be done without this change as the log_format clause has to be in the http block of the nginx configuration file, before the format is called with acccess_log further down the configuration file.


MooBaloo:
Also worth considering adding a new include clause for a new custom .conf file, like http_top.conf. Would avoid potentially creating issues with existing configurations using http.conf.

jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-849`

PR #850, opened 2021-01-28
Add new custom http.conf above includes for NPM-generated files 

MooBaloo:
Added a new clause for custom http_top.conf above the include clauses for NPM-generated files. Allows for more flexibility with adding custom nginx .conf files to NPM

Use case: adding a configuration change needs to be present before other custom configuration files are called and reference configuration from the custom http_top.conf file.
Example: add a new log_format in http_top.conf, then referencing it in a access_log clause in server_proxy.conf.

Could also simply move the existing include for http.conf, but opted not to in case it breaks existing uses of http.conf.


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-850`

jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-850`

jc21:
I'm ok with this as there doesn't seem to be a nicer way of doing this yet. Definitely looking to improve this global configuration stuff in future

PR #862, opened 2021-02-03
Update README.md

beamop:


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-862`

jc21:
So you're just making an *opinionated* change to quotes for a docker-compose example? Hardly worth a PR for this.

Furthermore, the convention is to single quote strings unless there is an encapsulated variable in the string.

PR #903, opened 2021-02-18
Prevent activating nginx config after editing disabled proxy host

### LINKED ISSUE #290: Proxy host is still active altough status is disabled
**Describe the bug**
- If a proxy host is manually disabled and changes in the SSL config are made, the host is still active altough the status still says 'Disabled'
- What version of Nginx Proxy Manager is reported on the login page?
*v2.0.14*

**To Reproduce**
1. Add a proxy host.
2. Click on the three dots and click Disable
3. Click on the three dots and click Edit > SSL
4. Change configuration. For example enable HTTP/2 and click Save.
5. See error. Proxy host ist still accessible altough status is Disabled.

**Expected behavior**
The host should not be accessible.

**System**
- Ubuntu 18.04. 
- NPM on docker 19.03.5, build 633a0ea838
- docker-compose version 1.25.3,

ogarai:
Check if proxy host is enabled and add an Nginx configuration only if the host is enabled to make sure it is consistent with the the db entry and the GUI

Fixes #290

### DIFF
--- a/backend/internal/proxy-host.js
+++ b/backend/internal/proxy-host.js
@@ -189,6 +189,10 @@ const internalProxyHost = {
 					expand: ['owner', 'certificate', 'access_list.[clients,items]']
 				})
 					.then((row) => {
+						if (!row.enabled) {
+							// No need to add nginx config if host is disabled
+							return row;
+						}
 						// Configure nginx
 						return internalNginx.configure(proxyHostModel, 'proxy_host', row)
 							.then((new_meta) => {


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-903`

jc21:
Docker Image for build 6 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-903`

PR #943, opened 2021-03-11
Update README.md

phantomski77:
Secrets need both global and local per-service declaration, otherwise fails as "not found". Local configuration added for both app and db.

### DIFF
--- a/docs/advanced-config/README.md
+++ b/docs/advanced-config/README.md
@@ -92,6 +92,8 @@ services:
     volumes:
       - ./data:/data
       - ./letsencrypt:/etc/letsencrypt
+    secrets:
+      - MYSQL_PWD
     depends_on:
       - db
   db:
@@ -106,6 +108,9 @@ services:
       MYSQL_PASSWORD__FILE: /run/secrets/MYSQL_PWD
     volumes:
       - ./data/mysql:/var/lib/mysql
+    secrets:
+      - DB_ROOT_PWD
+      - MYSQL_PWD
 ```
 
 


phantomski77:
**More context added:**

When I try to use the secrets example as published using `docker-compose up -d`, the resulting containers don't have an access to file stored secrets. This results in these errors in the container log:
```
[secret-init] Evaluating MYSQL_ROOT_PASSWORD__FILE,
[secret-init] ERROR: Cannot find secret in MYSQL_ROOT_PASSWORD__FILE,
[secret-init] Evaluating MYSQL_PASSWORD__FILE,
[secret-init] ERROR: Cannot find secret in MYSQL_PASSWORD__FILE
```
Looking at the Docker [Compose reference documentation (V3+)](https://docs.docker.com/compose/compose-file/compose-file-v3/#secrets), it says:
> Defining a secret does not imply granting a service access to it.

The examples that follow in the reference documentation all use a _per service_ secret access configuration (in other words _this secret will be used here in this container_) in addition to _top-level_ secrets configuration (in other words _here we declare the secret by linking it to an external file_),

When per-service configuration is added as per my pull request, the secrets are correctly evaluated, as demonstrated by:
```
[i] pre-init.d - processing /scripts/pre-init.d/01_secret-init.sh,
[secret-init] Evaluating MYSQL_ROOT_PASSWORD__FILE,
[secret-init] Success! MYSQL_ROOT_PASSWORD set from MYSQL_ROOT_PASSWORD__FILE,
[secret-init] Evaluating MYSQL_PASSWORD__FILE,
[secret-init] Success! MYSQL_PASSWORD set from MYSQL_PASSWORD__FILE
```
**Prerequisities:**
docker-compose version 1.25.0
docker version 19.03.13
docker-compose.yml using v3.7

jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-943`

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-943`

PR #1017, opened 2021-04-13
Apply 'Force SSL' and 'Enable Websockets' to full host definition

Stealthii:
The template options for these two options have been moved out of the root location block, and into the server block, in order for them to apply to custom locations set up via the proxy manager. This matches up with default expectations of how these toggle options work (as they are defined in the frontend host definition rather than the location).

If override behaviour is desired, these options can still be overridden by the existing "custom options" support offered.

### DIFF
--- a/backend/templates/dead_host.conf
+++ b/backend/templates/dead_host.conf
@@ -5,6 +5,7 @@ server {
 {% include "_listen.conf" %}
 {% include "_certificates.conf" %}
 {% include "_hsts.conf" %}
+{% include "_forced_ssl.conf" %}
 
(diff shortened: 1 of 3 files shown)


jc21:
Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1017`

jc21:
Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1017`

Issue #1018, opened 2021-04-14
Problems in renewing certificates


trenetics:
**Describe the bug**
Im having some trouble renewing certs with CloudFlare DNS Token. If i try to renew my certs manually it works, but takes a couple of minutes and the UI times out.

Auto renewals fail:

```
today at 21:26:56  `QueryBuilder#allowEager` method is deprecated. You should use `allowGraph` instead. `allowEager` method will be removed in 3.0
today at 21:26:56  `QueryBuilder#eager` method is deprecated. You should use the `withGraphFetched` method instead. `eager` method will be removed in 3.0
today at 21:26:56  QueryBuilder#omit is deprecated. This method will be removed in version 3.0
today at 21:27:24  [4/14/2021] [7:27:24 PM] [SSL      ] › ✖  error     Error: Command failed: /usr/bin/certbot renew --non-interactive --quiet --config "/etc/letsencrypt.ini" --preferred-challenges "dns,http" --disable-hook-validation  
today at 21:27:24  Challenge failed for domain te55er4ct.trenetics.io
today at 21:27:24  Attempting to renew cert (npm-20) from /etc/letsencrypt/renewal/npm-20.conf produced an unexpected error: Some challenges have failed.. Skipping.
today at 21:27:24  Challenge failed for domain dr1v3.trenetics.io
today at 21:27:24  Attempting to renew cert (npm-21) from /etc/letsencrypt/renewal/npm-21.conf produced an unexpected error: Some challenges have failed.. Skipping.
today at 21:27:24  Challenge failed for domain qu4n7um.trenetics.io
today at 21:27:24  Attempting to renew cert (npm-22) from /etc/letsencrypt/renewal/npm-22.conf produced an unexpected error: Some challenges have failed.. Skipping.
today at 21:27:24  Challenge failed for domain tr3nw1k1.trenetics.io
today at 21:27:24  Attempting to renew cert (npm-30) from /etc/letsencrypt/renewal/npm-30.conf produced an unexpected error: Some challenges have failed.. Skipping.
today at 21:27:24  Challenge failed for domain b1tw4rd3n.trenetics.io
today at 21:27:24  Attempting to renew cert (npm-34) from /etc/letsencrypt/renewal/npm-34.conf produced an unexpected error: Some challenges have failed.. Skipping.
today at 21:27:24  Challenge failed for domain a1rs0n1c.trenetics.io
[...]

PR #1022, opened 2021-04-16
install certbot from pip

jonasled:
It looks like the certbot version shipped with alpine is relativly old and has a few problems since a few days. It would be better if we install certbot through pip. Then we'll always get the newest version.

The problem is already discussed here: #1018


jc21:

jc21:
As the docker image has changed from Alpine to Debian now, this PR needs to be resolved manually.

jonasled:
I think with debian is a newer version shipped, so I can close this.

Issue #1033, opened 2021-04-23
SSL Options are not visible with a long Domain-List


ITbyHF:
I have one Certificate, that contains nearly all of my Domains, which was issued with the Cloudflare DNS-Challenge.
The upper 2 Options (Force SSL & HTTP/2 Support) in the SSL Menu are not visible when using a Certificate with a long domain list/name (see screenshot)
Version 2.8.0

Expected behavior
All Options visible or even an Option to give Certificates custom Names

![image](https://user-images.githubusercontent.com/61453797/115899069-9b0af600-a45e-11eb-8b9e-5dc79b6608d0.png)

Operating System
Docker-Compose / Ubuntu 20 (KVM VM)

jc21:
PR has been merged. Will be part of the next release.

PR #1034, opened 2021-04-23
don't fix the select height, to fix multiline select

jonasled:
Remove the fixed height of the div below the select. This should fix the bug mentioned in #1033


jc21:

jc21:
Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1034`

Issue #1015, opened 2021-04-13
invalid IPv6 address in resolver, scoped address, zone index 


JonasVgt:
I have made a Comment on a different Bug Report #149 but noticed later, that my problem was different to the one described there. That is why I made this separate Bug report.

### The Problem
I am running this Docker Image on my Rapsberry Pi Home server. When I try to start the Container, it is marked as 'unhealthy', the web-interface will not start and the following error is repeatedly being logged:
```
nginx: [emerg] invalid IPv6 address in resolver "[fe80::abd:43ff:feca:7350%2]" in /etc/nginx/conf.d/include/resolvers.conf:1
finish: applet not found
```

### Cause
When the Container is beeing started it will copy `/etc/resolv.conf` from the host machine (outside of the container) into the Container to the same location. The Content of that file will then be converted by the script located at ` nginx-proxy-manager/docker/rootfs/etc/services.d/nginx/run` of this repository in the following line:
```
# Dynamically generate resolvers file, if resolver is IPv6, enclose in `[]`
# thanks @tfmm
echo resolver "$(awk 'BEGIN{ORS=" "} $1=="nameserver" {print ($2 ~ ":")? "["$2"]": $2}' /etc/resolv.conf);" > /etc/nginx/conf.d/include/resolvers.conf
```
The output will then be written into `/etc/nginx/conf.d/include/resolvers.conf`. This file is read by nginx which will throw the error.

/etc/resolv.conf:
```
nameserver 8.8.8.8
nameserver fd00::3a10:d5ff:fe77:c05d
nameserver fe80::abd:43ff:feca:7350%2
search jonas.local
```
/etc/nginx/conf.d/include/resolvers.conf:
`resolver 127.0.0.11 [fd00::3a10:d5ff:fe77:c05d] [fe80::abd:43ff:feca:7350%2] ;`

_Note: `fd00::3a10:d5ff:fe77:c05d` is the IP of the Router which is connected to the Internet and directly to my Server via ethernet. `fe80::abd:43ff:feca:7350` is the IP of a second Router which is only responsible for the WLAN in my Network_  

The Error emerges, because one of the IPs is ending with the [ zone index](https://en.wikipedia.org/wiki/IPv6_address#Scoped_literal_IPv6_addresses_(with_zone_index)) `%2`. 

### How  to reproduce
[...]

PR #1041, opened 2021-04-26
Updated resolver-config conversion

JonasVgt:
Added "%" as a field separator in the conversion from resolv.conf to resolvers.conf.

Nginx doesn't allow zone-indices and will throw the following error
```
nginx: [emerg] invalid IPv6 address in resolver "[xxxx::xxxx:xxxx:xxxx:xxxx%2]" in /etc/nginx/conf.d/include/resolvers.conf:1
finish: applet not found
```
This will remove the zone index of link-local addresses and fixes the issue #1015.
More details in issue #1015


jc21:

jc21:
Not sure if you're still working on this? The PR is still in Draft status.

JonasVgt:
Sadly I can't reproduce the error anymore. Hence I will close this PR.

PR #1100, opened 2021-05-16
Displays entity id for each entry in frontend

chaptergy:
In this PR the entity ids are displayed inside the three-dot-menu on the right:

image

This is for all entities which are displayed in a table:

- Certificates
- Proxy Hosts
- 404 Hosts
- Redirection Hosts
- Streams
- Users
- Access Lists

Though the last two are not that relevant it probably won't hurt either.

Resolves https://github.com/jc21/nginx-proxy-manager/issues/174


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1100`

PR #1186, opened 2021-06-18
Use the certbot command from the PATH variable

chaptergy:
To fix https://github.com/jc21/nginx-proxy-manager/issues/1109, the path to the certbot command should not be hard-coded, but instead be taken from $PATH.  
Since it is only used inside `certificate.js` it is easy to validate. The certbot command is only used as a paramerter for `utils.exec()`, which in turn uses `child_process.exec()`, which does look up the commands in PATH.


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1186`

jc21:
CI does not validate certbot at all so confidence is required with some manual testing of the image

PR #1384, opened 2021-09-07
Add SSL and HTTP2 into IPv6 on listen.conf

bergi9:
I can only serve contents with IPv6 because I'm sitting behind CGN on IPv4. When enabling HTTP2 it still not serve contents with HTTP2 as there are missing arguments in the `listen`. But it still does the SSL encryption.
Previous to this commit it generates:
```
listen 80;
listen [::]:80;

listen 443 ssl http2;
listen [::]:443;
```
Now it generates:
```
listen 80;
listen [::]:80;

listen 443 ssl http2;
listen [::]:443 ssl http2;
```

### DIFF
--- a/backend/templates/_listen.conf
+++ b/backend/templates/_listen.conf
@@ -7,7 +7,7 @@
 {% if certificate -%}
   listen 443 ssl{% if http2_support %} http2{% endif %};
 {% if ipv6 -%}
-  listen [::]:443;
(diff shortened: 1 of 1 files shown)


jc21:
Ah good catch :)

jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1384`

PR #1474, opened 2021-10-09
Add DNS provider for Oracle Cloud Infrastructure (OCI) DNS

Djelibeybi:
Signed-off-by: Avi Miller <avi.miller@oracle.com>


Djelibeybi:
Tested `jc21/nginx-proxy-manager:github-pr-1474` on both `amd64` and `arm64` and both are working as expected.

PR #1493, opened 2021-10-17
Adds Portuguese translations

luandro:
We have a big need for this to be translated to Portuguese, as we intend various rural communities, with users new to the digital world, to be able to use this on their local networks.

I've started by adding translations to the "i18n/messages.json" file. But there's nowhere in the UI for language to be changed, and I didn't find in any part of the code translations to be changed based on browser language.

How can we start the process of adding other languages to the project?


jc21 (changes requested):
Thanks for the contribution! Just one comment as to the failure.

jc21:
FWIW this is the error CI is reporting:
```
07:57:55  ERROR in ./js/i18n/messages.json
07:57:55  Module build failed (from ./node_modules/messageformat-loader/index.js):
07:57:55  SyntaxError: Expected "#", "{", doubled apostrophe, end of input, escaped string, or plain char but "}" found.
```
However, adding translations to v2 isn't really practical since there's no UI switcher for languages and it's hard coded to english, despite the underlying support for language files.

PR #1559, opened 2021-11-04
Fixes custom location with capture groups

chaptergy:
- Renames `$upstream` to `$targetUri` in ngix config templates to prevent collision
- Does not append `$request_uri` to the url of a location if the location contains an opening bracket `(`, a closing bracket `(` and a tilde `~` signaling a regex capture group
- Does not append `$request_uri` if it is basically empty, so just a slash `/` as this would break domain-only requests (requests without a path)

Fixes https://github.com/jc21/nginx-proxy-manager/issues/1319

### DIFF
--- a/backend/templates/_location.conf
+++ b/backend/templates/_location.conf
@@ -1,5 +1,10 @@
   location {{ path }} {
-    set              $targetUri {{ forward_scheme }}://{{ forward_host }}:{{ forward_port }}{{ forward_path }}{% unless path contains "(" %}$request_uri{% endunless %};
+    set              $targetUri {{ forward_scheme }}://{{ forward_host }}:{{ forward_port }}{{ forward_path }};
+    {% unless path contains "~" and path contains "(" and path contains ")" %}
+    if ($request_uri != /){
+      set            $targetUri $targetUri$request_uri;
+    }
+    {% endunless %}
     proxy_set_header Host $host;
     proxy_set_header X-Forwarded-Scheme $scheme;
     proxy_set_header X-Forwarded-Proto  $scheme;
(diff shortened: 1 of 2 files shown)


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1559`

rainman369:
Hi - sorry to be "that person", but I'm not well versed in github or linux and just use unRAID + dockers to manage my media server. Do I have to do anything special to get this issue corrected or will this ultimately be an update that gets pushed out automatically? Thank you

chaptergy:
@rainman369 See https://nginxproxymanager.com/upgrading. But since you use unraid, it might be different, I have no idea how unraid works. Maybe something like [this](https://github.com/jlesage/docker-nginx-proxy-manager#unraid)? But first you have to wait until this is merged and a new release has been published. You can use the _watch_ button at the top of the github page and select _custom_ to select notifications only on new releases.

rainman369:
> @rainman369 See https://nginxproxymanager.com/upgrading. But since you use unraid, it might be different, I have no idea how unraid works. Maybe something like [this](https://github.com/jlesage/docker-nginx-proxy-manager#unraid)? But first you have to wait until this is merged and a new release has been published. You can use the _watch_ button at the top of the github page and select _custom_ to select notifications only on new releases.

Ah yes, I guess waiting for the release to actually come out would help :D Thank you!

jc21:

Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1559`

causefx:
@chaptergy I think you inadvertently caused the same issue by naming both location and proxy variables the same variable.  Those need to be different...  I would suggest keeping `$targetUri` for one of those and changing the other to whatever you want...  

Shall I open another issue up?

chaptergy:
@causefx I am assuming that variables _should_ be scoped to their respective location block, however there seems to be no documentation on that. But since nginx has lots of weird little features it's entirely possible this is not the case.  
However in https://github.com/jc21/nginx-proxy-manager/pull/1578 all this mess with the variables has been reverted anyways, since it's not working as intended either way due to other nginx limitations.

causefx:
yeah i hear ya man. either way no worries :)

PR #1565, opened 2021-11-04
Removes HSTS from admin ui

chaptergy:
Removes HSTS header from admin ui (port 81). Since it is not ssl encrypted by default and it is not possible to natively encrypt it this should not be there. This should be settable in a proxy host created to proxy the admin ui, otherwise the header would always be set, even if the proxy host does not have it set.

Fixes https://github.com/jc21/nginx-proxy-manager/issues/1560

### DIFF
--- a/backend/app.js
+++ b/backend/app.js
@@ -40,7 +40,6 @@ app.use(function (req, res, next) {
 	}
 
 	res.set({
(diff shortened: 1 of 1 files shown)


jc21:

Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1565`

PR #1578, opened 2021-11-09
Reverts back to proxy_pass without variables

chaptergy:
So apparently using the variable for the `proxy_pass` directive comes with a whole load of issues which are not easily fixable, and break how locations are expected to work. This reverts the template back to the normal `proxy_pass` usage without storing the target in a variable.


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1578`

PR #1614, opened 2021-11-29
Feature: Add two new headers to proxy.conf

### LINKED ISSUE #1609: proxy_set_header X-Forwarded-Host is missing in conf.d/proxy.conf
I would highly encourage to include the header `X-Forwarded-Host` in the default `proxy.conf` file.

This proxy header is used to let servers now what domain they are being proxied for and even required by some applications like [Authelia](https://www.authelia.com) (their provided [example conf](https://www.authelia.com/docs/deployment/supported-proxies/nginx.html#proxyconf) for Nginx).

If you want to read more, head to the documentation from mozilla: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host

Workaround with the current version would be to insert an advanced configuration in the web ui:
```
location / {
  # Proxy!
  include conf.d/include/proxy.conf;
  proxy_set_header  X-Forwarded-Host $host;
}
```

the1ts:
Fixes #1609. Adding both  X-Forwarded-Host  and X-Forwarded-Port, this is vital for some services behind a proxy (used to allow creation of absolute links in html). I've had to include at least the Host version in the past for Jenkins and Nexus.
Been running locally for 24 hours, does not appear to break any of my 15+ services currently running behind NPM would allow people to host those problem services without the need for advanced configuration.


the1ts:
This appears to be failing due to lack of python during the sqlite3 build in the backed.
```
[2021-11-29T13:53:52.222Z] make: Entering directory '/app/node_modules/sqlite3/build'
[2021-11-29T13:53:52.222Z]   ACTION deps_sqlite3_gyp_action_before_build_target_unpack_sqlite_dep Release/obj/gen/sqlite-autoconf-3310100/sqlite3.c
[2021-11-29T13:53:52.222Z] /bin/sh: 1: python: not found
```
This appears to be looking for python2 (/usr/bin/python) rather than python3 (/usr/bin/python3), anything changed in that regard in the CI system lately?

the1ts:
I think this is because node:latest is now based on Debian bullseye (SHA1 hashes for latest and current-bullseye match). Bullseye by default no longer has python 2 installed. So either we `apt-get update ; apt-get install -y python2.7` as part of that build `docker run` or use `node:current-buster` as we did before `node:latest` moved. 
If we do option 2 we are waiting until node-gyp and dependencies move to python 3 and hoping node support buster until that happens.

operinko:
Wasn't the EOL for Python 2.7 in 2020 already? Would make absolutely *no* sense to use that going forwards.

the1ts:
If node-gyp does support python 3 , then we need to fix the bullseye alternatives to allow python 3 to be used for /usr/bin/python so the build scripts will find python in their path.
From a new minimal bullseye install.
```
# python
-bash: python: command not found
# update-alternatives --install /usr/bin/python python /usr/bin/python3 2
update-alternatives: using /usr/bin/python3 to provide /usr/bin/python (python) in auto mode
# python --version
Python 3.9.2
```

jc21:
You might have to rebase on the `develop` branch as this python fix has been added

jc21:

Docker Image for build 3 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1614`

jc21:
I have to revert this PR due to #1717. Perhaps there is another way around this.

Re-reading the PR comment above, I use jenkins behind NPM and haven't had a problem with it in the past.

PR #1697, opened 2021-12-29
Sets certbot official dns plugin version requirement to match certbot version

chaptergy:
Currently official certbot dns plugins have no version requirement. This could lead to a newer version of a plugin being installed than the certbot has, and thus they could become incompatible.

This should fix https://github.com/jc21/nginx-proxy-manager/issues/1696.


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1697`

PR #1765, opened 2022-01-16
Set RSA key size to 4096 (defaults to 2048 which is debatably insecure)

marinmo:
If nothing else is specified the RSA key requested from letsencrypt is 2048 bits. This is possibly insecure and 4096 bit keys should be a negligible difference in processing speed.

### DIFF
--- a/docker/rootfs/etc/letsencrypt.ini
+++ b/docker/rootfs/etc/letsencrypt.ini
@@ -4,3 +4,4 @@ webroot-path = /data/letsencrypt-acme-challenge
 key-type = ecdsa
 elliptic-curve = secp384r1
 preferred-chain = ISRG Root X1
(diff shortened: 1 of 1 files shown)


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1765`

marinmo:
> If nothing else is specified the RSA key requested from letsencrypt is 2048 bits. This is possibly insecure and 4096 bit keys should be a negligible difference in processing speed.
I am aware NPM defaults to using elliptic curve certificates, but RSA certificates are also issued for compatibility reasons. This ensures that those certificates are equally secure.

PR #1839, opened 2022-02-08
removing HSTS include from location as it is superfluous. 

ncraig-oc:
HSTS header is already added at the server level and re-adding the host header at the location prevents inheritance from the advanced config include.

Per the NGINX documentation,
"These directives are inherited from the previous configuration level if and only if there are no add_header directives defined on the current level."

### DIFF
--- a/backend/templates/proxy_host.conf
+++ b/backend/templates/proxy_host.conf
@@ -51,7 +51,6 @@ proxy_http_version 1.1;
 
     {% endif %}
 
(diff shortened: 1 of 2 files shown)


jc21:

Docker Image for build 1 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-1839`

PR #1964, opened 2022-03-30
Update certbot-dns-cpanel plugin to 0.4.0

akic100:
The new version supports token authentication.


FreddyF1977:
Just encountered this issue, had to make the change manually in the container. Would be great not to have to do it on every container update.

PR #1999, opened 2022-04-13
Streams Basic Load Balancing (Allows multiple Forwarding Hosts)

eedev:
Hello,

I have added the ability to add several IPs/Hostnames to a Stream which is then defined into an upstream in the configuration.

I wanted to use your project (thank you for this!) to manage a simplistic implementation of [Nginx's TCP and UDP Load Balancing](https://docs.nginx.com/nginx/admin-guide/load-balancer/tcp-udp-load-balancer/) that would suit my needs.

Maybe someone else would be interested and add more options.


hayou:
Do you have any news from this features ? 
Regards

PR #2104, opened 2022-06-09
📚 add arabic local for V3

ajf-sa:
Hi there ! 

thanks for all work, I hope you accept this RP

### DIFF
--- a/frontend/src/locale/src/ar.json
+++ b/frontend/src/locale/src/ar.json
@@ -369,10 +369,10 @@
 		"defaultMessage": "Github"
 	},
 	"footer.userguide": {
-		"defaultMessage": "User Guide"
+		"defaultMessage": "دليل المستخدم"
 	},
 	"form.cancel": {
-		"defaultMessage": "Cancel"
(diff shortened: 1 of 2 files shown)


frontend/src/locale/src/ar.json
```diff
@@ -0,0 +1,560 @@
+{
+	"access-lists.title": {
+		"defaultMessage": "قوائم الوصول"
+	},
+	"acmesh.dns_acmedns": {
```
jc21, line 5: It's worth noting that if a definition doesn't exist here it will fall back to the `en.json` value. For lang keys starting with `acmesh.` you don't need to add them here since they _should_ have the same value in each language.

jc21:
Looks good and thanks for contributing here. Just that one comment from me.

Since this is an unreleased-dev branch it's not greatly documented, however you can ensure that the language files are happy by running the following:

```bash
cd frontend
yarn install
yarn check-locales
```

The CI will run this and fail the PR if this script isn't happy.

PR #2111, opened 2022-06-15
Column "Comments" in "Proxy Hosts" list

stefou84:
Hi there,
I discover and use NPM for a few days and it's really great!
Just a small suggestion, in the "Proxy Hosts" list, I think it might be interesting to add a "Comments" column to write some notes. What do you think ?


jc21:
I'm not personally working on v2 of this codebase, so if you want to add it in an elegant way, go for it.

PR #2293, opened 2022-09-26
Update README.md

GogoFC:
Deprecated docker network syntax updated so it doesn't give error anymore.


jc21:
This is already fixed in #2355. I know your PR was first but I went backwards and merged that one first, sorry..

Issue #2333, opened 2022-10-19
`DISABLE_IPV6` should also set `ipv6=off` in the `resolvers.conf`


czocher:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
When running under podman on a dns-enabled network, the internal container name is resolved correctly by the container (tested using `ping {container-name}`), but is not resolved correctly by `nginx`. Adding `ipv6=off` to the `resolvers.conf` fixes the issue. In my opinion this setting should automatically be added when setting the `DISABLE_IPV6` environmental variable to `true`.

**Nginx Proxy Manager Version**
 v2.9.18 

**To Reproduce**
1. Create a podman container for `nginx-proxy-manager` with its own network (required for DNS resolving to be enabled by default) and the `DISABLE_IPV6` environmental variable set to `true`.
2. Create another container on the same network.
3. Use the container name to create a proxy setting.
4. Nginx doesn't resolve the container by name although the container is visible and pingable.
5. Add the `ipv6=off` setting to the the `/etc/nginx/conf.d/include/resolvers.conf` for the internal DNS.
6. Reload the configuration with `nginx -s reload`.
7. The problem disappears.

**Expected behavior**
Adding `DISABLE_IPV6='true'` to the container config should add the `ipv6=off` setting to the `resolvers.conf`.

**Operating System**
Fedora CoreOS 36

PR #2335, opened 2022-10-20
Fix DISABLE_IPV6 flag handling

Czocher:
The DISABLE_IPV6 flag did not turn off ipv6 DNS requests performed by nginx. This commit changes it and makes nginx-proxy-manager more compatible with podman.

Related to bug #2333.

### DIFF
--- a/docker/rootfs/etc/services.d/nginx/run
+++ b/docker/rootfs/etc/services.d/nginx/run
@@ -24,8 +24,12 @@ chown root /tmp/nginx
 
 # Dynamically generate resolvers file, if resolver is IPv6, enclose in `[]`
 # thanks @tfmm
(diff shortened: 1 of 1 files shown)


Czocher:
Thanks for the quick response, gonna test it later today and notify you about my results.

Czocher:
Tested the PR with both settings - with `DISABLE_IPV6=true` and without. The problem stops appearing as soon as `DISABLE_IPV6` is enabled. The `resolvers.conf` is edited correctly:
* When `DISABLE_IPV6='true'` the `ipv6=off` appears in the `resolvers.conf`,
* When `DISABLE_IPV6` is not provided, `ipv6=off` does not appear in `resolvers.conf`.

To summarize: this PR works correctly according to my tests and can be merged.

PR #2382, opened 2022-11-08
Updated dependencies. Removed old unsupported plugins. 

SamTV12345:
This pr solves multiple issues:
- old dependencies: Plugins like the babel-loader plugin are considered deprecated. 
- dev dependencies: Typescript things were put into the dev dependencies section in package.json
- solved incorrect dependencies hierarchy when executed with npm
- updated prettier to ignore line endings. Windows adds a carriage return line feed which prettier doesnt like. This is always removed when it is versioned into git.

A big improvment would be to switch from CRA which is really bloated to vite React Typescript plugin. The webserver starts in two-three seconds, I can do dynamic imports without writing everywhere lazy import and I don't download a lot of files when installing the project.


SamTV12345:
Force pushed signing key.

SamTV12345:
Will this pr ever be checked?

PR #2409, opened 2022-11-17
Feature: Use configurable PUID and PGID to get rid of permission issues

plantysnake:
So, I am using nginx-proxy-manager with my NAS and it is great.
Except that if you map volumes, like /data or /etc/letsencrypt, to some folders for more flexible configuration, they have different owner/group than all other files and folders.
This is solved by creating abc user and then using him as owner of the files.


plantysnake:
#2721 does everything from this PR, closing as not needed

Issue #2388, opened 2022-11-09
Add ability to configure "worker_processes"


PingTrip:
**Is your feature request related to a problem? Please describe.**
For a home network/low traffic deployment the current setting of `worker_processes auto;` is overkill since it spawns an nginx process for every core. 

**Describe the solution you'd like**
Provide a mechanism to override the current hard-coded "auto" setting.

**Describe alternatives you've considered**
I tried overriding the parameter via `/data/nginx/custom/root.conf` which resulted in an Nginx error:
`nginx: [emerg] "worker_processes" directive is duplicate in /data/nginx/custom/root.conf:1`.

k4my4b:
Exact same issue here.
An environment variable like `WORKER_PROCESSES` would suffice.
But while we're at it can we please add environment variables for [thread pools](https://www.nginx.com/blog/thread-pools-boost-performance-9x/) as well, thanks!

fevernova90:
Bump up! 

Using this in production of a small growing app. There's no direct way to override `worker_connections`. 1024 is surely not enough.

CreeperFace00:
Having this issue as well, would love to see this implemented!

msycn:
same here please help us to change worker_connections

msycn:
> @msycn That is already possible thanks to the commit @Blooym made. Just put that line in and it will work just fine.`/data/nginx/custom/events.conf`

Wow ,you are my hero ! Thanks for your quick reply!!!

ansred:
@kroese getting the same issue, as reported here:

https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3345

Any idea what is that line for?
`/data/nginx/custom/events.conf`

Do we add it as env in the docker compose? or is it something else?

ansred:
@kroese Here is my docker compose

```
version: "3"

services:
  app:
    image: 'jc21/nginx-proxy-manager:latest'
    container_name: Nginx_Proxy_Manager_SQLite
    restart: always
    ports:
      - '80:80'        # Public HTTP Port:
      - '443:443'      # Public HTTPS Port:
      - '81:81'        # Admin Web Port:
    environment:
        DB_SQLITE_FILE: "/data/database.sqlite"
        DISABLE_IPV6: 'true'

    volumes:
      - /volume1/docker/containers/npm/config/data:/data
      - /volume1/docker/containers/npm/config/cert:/etc/letsencrypt

```

[...]

Issue #1912, opened 2022-03-05
How to modify the events context in nginx.conf?


jearton:
**Is your feature request related to a problem? Please describe.**

```bash
tail -f fallback_error.log
```

```bash
2022/03/05 12:45:07 [warn] 415#415: 1024 worker_connections are not enough, reusing connections
2022/03/05 12:45:07 [alert] 415#415: 1024 worker_connections are not enough
```

**Describe the solution you'd like**

Modify to this: 

```nginxconf
events {
    multi_accept       on;
    worker_connections 65535;
}
```

**Describe alternatives you've considered**

**Additional context**

Geczy:
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/2388#issuecomment-1807353881

PR #2444, opened 2022-12-02
Load events configuration from custom file

### LINKED ISSUE #1435: worker_connections increase
Hi, I have nginx proxy manager running in docker container and it hosts wordpress. In peak time web sometimes throw "500 Internal Server Error".
image

When searching through the error logs I always found this:
```
[alert] 2190841#2190841: 1024 worker_connections are not enough
[warn] 2190841#2190841: 1024 worker_connections are not enough, reusing connections
```

I've wanted to increase worker_connections, but I was unlucky as the only possible way is to add it to "/data/nginx/custom/root.conf":
```
events {
    worker_connections  10024;
}
```
That resulted into this error: "events" directive is duplicate in /data/nginx/custom/root.conf:1

So I'm out of ideas and replacing /etc/nginx/nginx.conf directly seems sketchy to me. Is there another way?

Thank you.

BitsOfAByte:
## Changes

Instead of forcing users to mount a custom `nginx.conf` file to modify the **events** configuration, we should instead allow them to use a `/custom/events.conf` file to modify it, as is standard for other advanced parts of NPMs configuration. It also reverts the default `worker_connections` amount to NGINX's default to prevent duplicate key issues if a user wants to set a custom `worker_connections` amount.

## Rationale

I was consistently running into issues when using NPM as I was unable to modify my worker_connections amount without some hacky workarounds, so I wanted to make it as easy as changing other parts of the configuration.

## Additional Information

Let me know if anything in this PR needs adjusting, I am by no means an NGINX expert and just wanted to upstream some of my local changes. If this is unwanted, feel free to just close the PR.

## Linked Issues

Fixes #1435, #1912, #2388

### DIFF
--- a/docker/rootfs/etc/nginx/nginx.conf
+++ b/docker/rootfs/etc/nginx/nginx.conf
@@ -15,7 +15,7 @@ error_log /data/logs/fallback_error.log warn;
 include /etc/nginx/modules/*.conf;
 
 events {
-	worker_connections  1024;
+	include /data/nginx/custom/events[.]conf;
 }
 
 http {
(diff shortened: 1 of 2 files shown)


msycn:
Wow ,you are my hero ! Thanks for your idea ,wish will pull request soon

PR #2512, opened 2022-12-31
Use latest setuptools

dormancygrace:
I've checked setuptools 64 & 65 - all works as intended (tested on arm64)


binhotvn:
Why you remove those lines instead of change version only?

dormancygrace:
> Why you remove those lines instead of change version only?

Because there is no need to uninstall latest setuptools and then install version 58

jc21:
The downgrade was initially required because of failing pip installs for certbot plugins. The only way your change can be tested as working is to install every certbot DNS plugin the app supports.

dormancygrace:
> The downgrade was initially required because of failing pip installs for certbot plugins. The only way your change can be tested as working is to install every certbot DNS plugin the app supports.

I've tested reg.ru plugin, duckdns they work fine

PR #2530, opened 2023-01-10
ACL changes should not blow away cert config

### LINKED ISSUE #2254: wrong certificate being served after editing access list
**Checklist**
  - Yes 
  - Yes 
  - Yes 

**Describe the bug**
Sometimes, and I can't really pin down when, HTTPS sites added in NPM break. This can happen after a container-restart or during normal operation. When this happens, chrome reports either `ERR_HTTP2_PROTOCOL_ERROR ` or `NET::ERR_CERT_AUTHORITY_INVALID` with a certificate issued to and from localhost. Restarting the container or the host doesn't seem to fix this issue. 

Accessing the same URL over HTTP instead of HTTPS works perfectly fine.

The only reliable way to fix this seems to be to click Edit and then Save on each individual Proxy Host.

v2.9.18 

**Expected behavior**
NPM directs the traffic normally

**Screenshots**
image

**Operating System**
debian 11

jmerdich:
When editing an access list, all affected sites get their config updated without certificates, functionally breaking https on those sites until they get a manual config change that rewrites the config properly. Properly fetch the certificate from the DB before `bulkGenerateConfigs` so it gets updated right.

I am not certain whether there are similar bugs in other places that use `bulkGenerateConfigs`.

Should fix #2254

### DIFF
--- a/backend/internal/access-list.js
+++ b/backend/internal/access-list.js
@@ -218,7 +218,7 @@ const internalAccessList = {
 				// re-fetch with expansions
 				return internalAccessList.get(access, {
 					id:     data.id,
-					expand: ['owner', 'items', 'clients', 'proxy_hosts.access_list.[clients,items]']
+					expand: ['owner', 'items', 'clients', 'proxy_hosts.[certificate,access_list.[clients,items]]']
 				}, true /* <- skip masking */);
 			})
 			.then((row) => {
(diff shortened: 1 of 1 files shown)


niv:
I can confirm the change in this PR also seems to fix invalid/self-signed cert issues in my local setup after re/starting the docker stack.

PR #2582, opened 2023-02-03
Use ssl_reject_handshake to reject requests to default https site

TheBeeZee:
Instead of creating a dummy certificate, we can return an SSL protocol error, which will generate a descriptive error message in the browser.


TheBeeZee:
Before:

image


After: 
image

PR #2606, opened 2023-02-11
Changes log filenames from ID to (first) domain

### LINKED ISSUE #746: Change access log filenames to domain name (or add option)
**Is your feature request related to a problem? Please describe.**
It's very frustrating to try and determine which log file is for which site since they are all named after the arbitrary ID that they are assigned when created.

It would be nice to instead have them named after the first domain in the `domain_names` list.
```
~BEFORE~
# ls
proxy_host-1.log  proxy_host-10.log  proxy_host-11.log	proxy_host-2.log  proxy_host-3.log  proxy_host-4.log  proxy_host-5.log	proxy_host-6.log  proxy_host-7.log  proxy_host-8.log  proxy_host-9.log

~AFTER~
# ls
domain.example.com.log fqdn.something.log ...
```

Using the [template on how the domain names](https://github.com/jc21/nginx-proxy-manager/blob/030e5535495dc7a0050fc34f2b9aef76ba45c3d2/backend/templates/_listen.conf#L15) are set for `server_name`, I assume this can be accomplished by simply changing the access log line to something like this (proxy host as the example):
```
access_log /data/logs/{{ domain_names[0] }}.log proxy;
OR
access_log /data/logs/{{ domain_names | first }}.log proxy;
```

I tried to just add a custom `access_log` line to the advanced configuration, but instead of overwriting the previous setting it sends logs to both files.

**Additional context**
N/A

joewesch:
Closes: #746

This change would be slightly not backwards compatible as the log filenames would change for all new logs and all existing files would be abandoned and not pruned. Any guidance or assistance to add this as a setting in the UI or something rather than just change the default would be welcomed.


KGBist2000:
Hi Joewesch,
nice change. Did you checked what will happen, if more then one domain is entered into the gui by creating a new host ?

csantospminds:
I'm very interested in this, any idea if it's possible to change the log filename from ID to domain? 
Adding the option to read the logs in the GUI would also be cool because right now it's difficult to find a particular log for a dozen or more vhosts.

hirenshah:
Could this be expanded to name the conf files in a similar manner?

pXius:
Is this PR still being considered?

joewesch:
Pinging @jc21 to see if there is any feedback?

Issue #669, opened 2020-10-20
Resolver directive not working?


rlogiacco:
**Checklist**

- [x] Please read the [setup instructions](https://nginxproxymanager.com/setup/)
- [x] Please read the [FAQ](https://nginxproxymanager.com/faq/)

**What is troubling you?**

I'm using proxy manager to proxy docker containers but when a container is restarted the proxy manager reports a 502 BAD GATEWAY.
I tried adding the `resolver` directive in the advanced configuration, like `resolver 127.0.0.11 ipv6=off valid=30s;` but it doesn't seem to work:
![image](https://user-images.githubusercontent.com/199473/96600141-eba76800-12f0-11eb-9b37-03c0bdbaa948.png)

Do I need to use a variable for the resolve to work properly?

Elmardus:
This may be related to the issue #633

rlogiacco:
I managed to have it working and it also solves #633 : you must use a variable and the resolver directive. The variable prevents the DNS resolution at boot from nginx, the resolver determines how frequently such variable is checked against the internal docket DNS.

Wadera:
I would suggest to show full resolution (for any other people who may face similar problem) and close Issue then ;)

Elmardus:
@rlogiacco How did you use the variable and the resolver directive?
I found [this guide](https://sandro-keil.de/blog/let-nginx-start-if-upstream-host-is-unavailable-or-down/), but I have not figured out how to implement this in Nginx Proxy Manager.

chaptergy:
I'm pretty sure he means like this:

```
location / {
    resolver 127.0.0.11 valid=30s;
    set $custom_upstream example.com;
    proxy_pass http://$custom_upstream:80;
}
```

The `$custom_upstream` is the variable which will prevent DNS resolution at startup.

Which you can paste into your advanced tab.

kabadisha:
For anyone who finds this in future:
This is a valid workaround for the problem where Nginx Proxy Manager fails to start if it can't resolve on of the upstream servers (e.g. another docker container on your unRaid server that is not currently started).

This issue only occurs with Proxy Hosts in NPM that have custom locations defined. The workaround is to remove the custom location and instead declare it manually in the 'Advanced' tab of the host like so:
```
location / {
    resolver 127.0.0.11 valid=30s;
    set $custom_upstream example.com;
    proxy_pass http://$custom_upstream:80;
}
```
Just replace `example.com` with the host that is not always available.

Issue #633, opened 2020-10-05
Unavailable host in proxy configuration preventing startup of NPM


Elmardus:
**Describe the bug**
- When using a Forward Hostname in the Proxy settings, Nginx Proxy Manager fails to start if this host is unavailable.
- NPM version 2.5.0

**To Reproduce**
1. Setup a Nginx-proxy-manager proxy host with the hostname of a docker container as the _Forward Hostname_, example shown below:

![image](https://user-images.githubusercontent.com/25563589/95137987-7fadf700-0769-11eb-8f1a-b83e8336cce1.png)

2. Stop or remove the container with this hostname
3. Restart NPM

NPM will now fail to start, with the following error log:
```
❯ Enabling IPV6 in hosts: /etc/nginx/conf.d
  ❯ /etc/nginx/conf.d/default.conf
  ❯ /etc/nginx/conf.d/include/assets.conf
  ❯ /etc/nginx/conf.d/include/ip_ranges.conf
  ❯ /etc/nginx/conf.d/include/letsencrypt-acme-challenge.conf
  ❯ /etc/nginx/conf.d/include/proxy.conf
  ❯ /etc/nginx/conf.d/include/force-ssl.conf
  ❯ /etc/nginx/conf.d/include/ssl-ciphers.conf
  ❯ /etc/nginx/conf.d/include/block-exploits.conf
  ❯ /etc/nginx/conf.d/include/resolvers.conf
  ❯ /etc/nginx/conf.d/production.conf
❯ Enabling IPV6 in hosts: /data/nginx
  ❯ /data/nginx/default_host/site.conf
  ❯ /data/nginx/proxy_host/18.conf
  ❯ /data/nginx/proxy_host/28.conf
  ❯ /data/nginx/proxy_host/7.conf
  ❯ /data/nginx/proxy_host/32.conf
  ❯ /data/nginx/proxy_host/9.conf
  ❯ /data/nginx/proxy_host/24.conf
  ❯ /data/nginx/proxy_host/26.conf
  ❯ /data/nginx/proxy_host/3.conf
  ❯ /data/nginx/proxy_host/25.conf
  ❯ /data/nginx/proxy_host/22.conf
  ❯ /data/nginx/proxy_host/23.conf
  ❯ /data/nginx/proxy_host/31.conf
  ❯ /data/nginx/proxy_host/27.conf
  ❯ /data/nginx/proxy_host/5.conf
  ❯ /data/nginx/proxy_host/4.conf
[...]

PR #2672, opened 2023-03-12
FIX: Ngnix fails to start if custom location upstream is unavailable/unreachable

kabadisha:
:warning: **Potentially breaking change**:
Upstream services with special configurations to compensate for not being hosted at root of domain are affected (e.g. Sonarr & Radarr). See section **A potential problem** below for an explanation.

### Issue:
Currently, if a Proxy Host has a custom location and the host being forwarded to is down, Nginx fails to start.
See [here](https://github.com/NginxProxyManager/nginx-proxy-manager/discussions/2228), [here](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/633) and [here](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/669) for examples of others who have encountered this issue.

### Impact:
In this scenario:
- All other hosts managed by NPM are now also taken out.
- NPM cannot be started in order to disable or edit the host configuration in order to rectify the issue.

### Solution:
Declare a variable for `forward_host` instead of simply injecting it directly into the `proxy_pass` directive.

Although not strictly necessary to fix the issue, this PR also adds similar variables for `forward_scheme` and `forward_port` as this keeps things consistent and these variables could be useful in any custom elements an end-user might want to add.

Dialog

### Handling paths and GET params properly:
With further testing and reading I discovered that the Nginx `proxy_pass` directive does not gracefully handle paths when using variables ([source](https://dev.to/danielkun/nginx-everything-about-proxypass-2ona#let-nginx-start-even-when-not-all-upstream-hosts-are-available)):
>  Changing
> ```
> proxy_pass https://localhost:5000/api/;
> ```
> to
> ```
>set $upstream https://localhost:5000;
> proxy_pass $upstream/api/;
> ```
> you might think should result in exactly the same, you might be surprised. The former will hit your upstream server with /api/foo?bar=baz with our example request to /webapp/foo?bar=baz. The latter, however, will hit your upstream server with /api/. No foo. No bar. And no baz. :-(

I also found that @chaptergy had [attempted something similar in the past, but reverted it](https://github.com/NginxProxyManager/nginx-proxy-manager/commit/1f879f67a9bed8f5eb444d00ad47c353b8a75db8?diff=unified) because of this issue.
@chaptergy I believe the approach I have found works well. It certainly works in my testing. I'd appreciate your second opinion & testing though.

### Testing:
It can be a bit tricky testing this because it isn't always obvious exactly what request is being passed to the upstream.
For my testing, I used [http-https-echo](https://code.mendhak.com/docker-http-https-echo/) as my upstream as it simply shows the details of the request being passed through directly in the browser. Super handy.

### A potential problem:
For one of my proxy hosts, I had a custom location: `/sonarr` with the forward hostname set to `sonarr`.
It turns out that prior to this change, the `proxy_pass` directive was automatically including the custom location path `/sonarr` in the proxied request to the upstream Sonarr.
This meant that in Sonarr's configuration, I had configured the property `URL Base` to `/sonarr` to compensate.

With this PR, the `proxy_pass` directive now actually does what the NPM UI describes and so uses the path specified by the user in the `Forward Hostname / IP` field instead. If no path is provided there, then no path is used in the upstream request.
This means that with this change in place, I either have to set the `URL Base` property in Sonarr back to the default `/` or set the `Forward Hostname / IP` field in NPM to `sonarr/sonarr`.

This also affects Radarr in the same way.

### Conclusion:
Personally, I think this PR actually makes NPM behave in a way that is more flexible, explicit and in keeping with what the UI actually presents. However, I'm pretty sure that this makes it a potentially **breaking change** for many users who have 'worked around' this issue by configuring their upstream systems to recognise that they are not hosted at the root of the domain. It also needs some decent testing by the community.

**On the flip side, this PR would:**
- Make it less likely that less experienced users would get tripped up by NPM silently adding a path to upstream requests and make compensating configuration at the upstream service unnecessary.
- Enable compatibility with upstream services that are not able to be configured to operate at non-root paths.


kabadisha:
Having done a bit more reading, I think there may be some issues with this fix. Specifically, I suspect that it will not handle paths and request parameters correctly.
I'm going to do more testing.

kabadisha:
Ok, so I think I have found an approach that works.
I trim the location path off of the nginx `$request_uri` variable and use logic at the LiquidJS templating layer to determine whether or not slashes needed to be added. There are four different permutations depending on whether or not the user adds a trailing slash to the location path or the forwarding path.
The `$request_uri` variable includes both the requested path, but also any GET parameters and so this approach handles both.

StefaBa:
> Currently, if a Proxy Host has a custom location and the host being forwarded to is down, Nginx fails to start.

FYI: nginx also fails to start if the destination of a stream is a hostname. 
I'm assuming your fix only fixes the startup issue of proxy hosts with a hostname?

kabadisha:
> FYI: nginx also fails to start if the destination of a stream is a hostname.

I didn't know that. I've never played with the stream function. Perhaps I'll take a look.
> I'm assuming your fix only fixes the startup issue of proxy hosts with a hostname?

That's correct.

kabadisha:
Nuts. I've just found an upstream container in my setup where this change causes a redirect loop.
Investigating...

kabadisha:
Found a bug in the PR:
If both the location path lacks a trailing slash and there is no forward path specified, the regex match doesn't work as desired and so the `$uri_remainder` variable ends up as the location path.
It's late here and I'm tired, so going to pick this up another time. I have a feeling that there's a neater & more reliable way to trim the location path off the front of the `$request_uri` variable in Nginx.

jack-mil:
Hey, good to see this issue being worked on! I noticed the exact problem for a container that I occasionaly stop. Whats the best manual workaround for now?

nunntom:
I've just tested this PR and it works beautifully for my use case without any change to the configuration. 

I agree with the comment on the other thread that it's not desirable for all the hosts to go down if just one fails to start. I'd rather rely on uptime monitoring to notice that one of my sites is not working than to lose the whole lot in one go.

TheTrueRandom:
Would be great if this PR could be considered. It's a pain having 20 proxy targets while some can occasionally die which is ok. During runtime it will be a 503 which is perfectly fine. But on restart npm won't start and all 20 proxy targets are down. 

Using IP addresses is not a solution. It won't work good with docker or kubernetes for which npm should be designed for?

nz2o:
Echoing TheTrueRandom, every time there's a broken update, such as in NextCloud, it takes down everything else with it, if NPM cannot start when unable to find an upstream host to proxy.

MaciejGorczyca:
Would be great to be able to start NPM when some other containers are not on/present anymore.

kabadisha:
Hi all, I'm sorry I went quiet on this one. Life has been busy and so I haven't had a chance to pick this back up.

I will try and make some time to look at improving this PR to a point where I am satisfied with it.

xrh0905 (approved):

kabadisha:
@xrh0905 Thanks for approving the change however, this PR still has some bugs so I don't think it's ready to submit for review. Life has been busy recently so I haven't had a chance to work on this PR.

bmerigan:
I heard this issue has actually been fixed in Niginx Plus.

chefkoch:
Bump to keep it open. Still a problem here on my side.

maxiwheat:
Just found out this issue, I'm having the same behavior (npm won't start if upstream is unavailable/unreachable, but even if the unavailable container doesn't have a custom location.

Giraffaman:
just encountered this while testing netmaker which requires a custom location to be defined - when I stopped the netmaker containers after testing I ran into this issue and found #2228 - defining the custom location in the "advanced" tab as described in https://github.com/NginxProxyManager/nginx-proxy-manager/discussions/2228#discussioncomment-5279030 now allows npm to start, even if my netmaker stack is down. Haven't found any other issues with this workaround as of yet.

xrh0905:
.

Giraffaman:
is there any particular reason for not merging this?

kabadisha:
Speaking as the original author of this PR, I think it probably needs more testing of edge-cases before it's safe to merge.
I'm not an nginx specialist and life took over before I could make time to really push this over the line.
I've had two kids since I raised this PR, so as much as I woul like to, I'm probably going to struggle to make sufficient time to get this it in really good shape and even then, it would still need some input from one of the maintainers to check my work.

Stephan-P:
My NPM fell over today. It wouldn't restart.
Checking its log I noticed a reference to a container name that I had put DOWN earlier today.
Who would have thought that a reference to container name that is NOT UP would be a blockade in starting NPM?

I try to bridge my containers in an external network 'web-proxy' and reference them in NPM by name + port.
Maybe I should stop doing that and always reference the forwarding by IP  + port

jonofmac:
Agree that this PR really should be pulled in as default behavior or at the very least be CONFIGURABLE. It is ABSOLUTELY insane that the entire proxy server fails to start if ANY of the proxies is unresolvable. I have 33 proxy services and sometimes they go down, which I catch with another service. Yet i cannot START npm when this happens. This is wild behavior.

"Yeah 32 of my 33 services are fine, but 1 of them i can't resolve, so instead of a 502 or other error if someone tries to reach that proxy, I'm just going to fail all 33 services and not provide even a web interface to log in and disable the service manually. No. Instead you have to ssh into the server, find the config file and remove it in order to get NPM to boot"

Issue #2632, opened 2023-02-25
Let's Encrypt wildcard certificate for reg.ru issue


navhawk80:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes 
- Are you sure you're not using someone else's docker image?
  - Yes 
- Have you searched for similar issues (both open and closed)?
  - Yes 

**Describe the bug**
When I try to get Let's Encrypt wildcard certificate for reg.ru I get error:
"certbot: error: unrecognized arguments: --certbot-regru:dns-credentials /etc/letsencrypt/credentials/credentials-..." in log.....

**Nginx Proxy Manager Version**
v2.9.19

habibullinrsh:
I haven't been able to solve this problem yet.

tarkh:
Hi guys! Same error with `reg.ru`. For now NPM emitting this command on SSL creation with reg.ru DNS plugin selected:

`mkdir -p /etc/letsencrypt/credentials 2> /dev/null; echo 'certbot_regru:dns_username=myuser
certbot_regru:dns_password=mypassword' > '/etc/letsencrypt/credentials/credentials-1' && chmod 600 '/etc/letsencrypt/credentials/credentials-1' && pip install certbot-regru~=1.0.2  && certbot certonly --config "/etc/letsencrypt.ini" --cert-name "npm-1" --agree-tos --email "my@email.com" --domains "*.mydomain.com" --authenticator certbot-regru:dns --certbot-regru:dns-credentials "/etc/letsencrypt/credentials/credentials-1"`

But this is incorrect, because we don't need prefix `certbot_regru:`. Correct command is:

`mkdir -p /etc/letsencrypt/credentials 2> /dev/null; echo 'dns_username=myuser
dns_password=mypassword' > '/etc/letsencrypt/credentials/credentials-1' && chmod 600 '/etc/letsencrypt/credentials/credentials-2' && pip install certbot-regru~=1.0.2  && certbot certonly --config "/etc/letsencrypt.ini" --cert-name "npm-1" --agree-tos --email "my@email.com" --domains "*.mydomain.com" --authenticator dns --dns-credentials "/etc/letsencrypt/credentials/credentials-1"`

Any ideas for hot-fix?

PR #2674, opened 2023-03-12
Fixed full_plugin_name for reg.ru plugin

tarkh:
For `reg.ru` plugin we don't need prefix `certbot_regru:` in it's `credentials` and `full_plugin_name`.
#2632


dormancygrace:
https://github.com/NginxProxyManager/nginx-proxy-manager/pull/2702#issuecomment-1474589327

same, but with build =)

PR #2714, opened 2023-03-19
FIX: Ngnix fails to start if upstream host for stream is unavailable/unreachable

kabadisha:
### Issue:
Currently, if the upstream host being forwarded to is down, Nginx fails to start.
Reported by @StefaBa here: https://github.com/NginxProxyManager/nginx-proxy-manager/pull/2672#issuecomment-1475361687

### Impact:
In this scenario:
- All other hosts managed by NPM are now also taken out.
- NPM cannot be started in order to disable or edit the host configuration in order to rectify the issue.

### Solution:
Declare a variable for `forwarding_host` instead of simply injecting it directly into the `proxy_pass` directive.
Use variables for `proxy_pass` directive in order to prevent startup failure if the upstream host is down.

@StefaBa this one is for you. I think it should work, but I don't use this functionality, so can you please test it?

### DIFF
--- a/backend/templates/stream.conf
+++ b/backend/templates/stream.conf
@@ -3,35 +3,37 @@
 # ------------------------------------------------------------
 
 {% if enabled %}
-{% if tcp_forwarding == 1 or tcp_forwarding == true -%}
-server {
-  listen {{ incoming_port }};
-{% if ipv6 -%}
-  listen [::]:{{ incoming_port }};
-{% else -%}
-  #listen [::]:{{ incoming_port }};
-{% endif %}
+  set $forwarding_host {{ forwarding_host }};
+  set $forwarding_port {{ forwarding_port }};
+  {% if tcp_forwarding == 1 or tcp_forwarding == true -%}
+    server {
+      listen {{ incoming_port }};
+    {% if ipv6 -%}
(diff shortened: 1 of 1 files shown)


kabadisha:
FYI: There are actually only a couple of functionally changed lines. Most of the change is indentation which I added to make the code easier for me to read.

StefaBa:
npm still fails to start, here's a pastebin of the startup log: https://pastebin.com/XW8z4GLK

fwiw: newhost.wan.vm is currently not only offline, but also unknown to the local DNS server.
-> nslookup newhost.wan.vm currently yields:
> ** server can't find newhost.wan.vm: NXDOMAIN

with newhost.wan.vm being offline, but known to the local DNS server NPM does come online (both with the current :latest version and with :github-pr-2714)

kabadisha:
Can you share what `3.conf` looks like?

StefaBa:
here you go: https://pastebin.com/eufsKe7W

i since tried recreating this stream in :github-pr-2714, but that always shows the stream as "Offline" in the web-ui, even when newhost.wan.vm is online.

kabadisha:
Sorry, I should have been more specific - I wanted to see `3.conf` generated by `:github-pr-2714` to make sure I didn't screw something up.

PR #2721, opened 2023-03-21
Docker users and groups, refactor configuration

jc21:
Adds support to run processes as a user/group, defined with PUID and PGID environment variables
- Detects if image is run with a user in docker command and fails if so
- Adds s6 prepare scripts for adding a 'npmuser'
- Split up and refactor the s6 prepare scripts
- Runs nginx and backend node as 'npmuser'
- Changes ownership of files required at startup

Refactor configuration
- No longer use config npm package
- Prefer config from env vars, though still has support for config file
- No longer writes a config file for database config
- Writes keys to a new file in /data folder
- Removes a lot of cruft and improves config understanding


plantysnake:
Hello, I just want to point out permission issue with logrotate:
`error: skipping "/data/logs/proxy-host-9_error.log" because parent directory has insecure permissions (It's world writable or writable by group which is not "root") Set "su" directive in config file to tell logrotate which user/group should be used for rotation.`

Solution would be changing this
```
/data/logs/*_access.log /data/logs/*/access.log {
    create 0644 root root
    weekly
    rotate 4
    missingok
	@@ -12,7 +13,8 @@
}

/data/logs/*_error.log /data/logs/*/error.log {
    create 0644 root root
    weekly
    rotate 10
    missingok
...
```
to this
```
/data/logs/*_access.log /data/logs/*/access.log {
    su root root
    create 0644 npmuser npmuser
    weekly
    rotate 4
    missingok
	@@ -12,7 +13,8 @@
}

/data/logs/*_error.log /data/logs/*/error.log {
    su root root
    create 0644 npmuser npmuser
    weekly
    rotate 10
    missingok
....
```
Or something like that.

And as this PR is more complex, #2409 should be closed after this one is merged.

PR #2816, opened 2023-04-10
Allow to include custom configuration in the location / of proxy blocks

jebabin:
This allow to customise the location / of proxy blocks

### DIFF
--- a/backend/templates/proxy_host.conf
+++ b/backend/templates/proxy_host.conf
@@ -41,6 +41,8 @@ proxy_http_version 1.1;
 
     # Proxy!
     include conf.d/include/proxy.conf;
+    # Custom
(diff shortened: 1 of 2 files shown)


webysther:
I think the #2899 is a complimentary to this one.

jebabin:
> I think the #2899 is a complimentary to this one.

Unless I misunderstand the {{id}} things, It’s different as your patch allow a custom file per proxy entry while mine allow to have a custom config global to all proxy entries. 
custom config per proxy can be done via GUI.

My use case is for example to customize the logging with if conditions (which can’t be done at an higher level of config) globally for all proxies

m-thalmann:
I would need this too! Any plans of merging this @maintainers?

Issue #2011, opened 2022-04-19
SSL certificate error


ahmedelemamn:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
i have a fresh NPM image running and tried to generate SSL certificate for my domain
i tried both http/dns challenges
for http challenge i get this error:

```
Communication with the API failed, is NPM running correctly?
```

or this one:

```
example.example.com: There is no server available at this domain. Please make sure your domain exists and points to the IP where your NPM instance is running and if necessary port 80 is forwarded in your router.
```
for the second error i made sure my DNS record is configured as DNS only and not proxied on cloudflare and i have both port 80 and 443 forwarded on my WAN router

if i opted for DNS challenge i get this error
```
Error: Command failed: certbot certonly --config "/etc/letsencrypt.ini" --cert-name "npm-3" --agree-tos --email "xxxx@gmail.com" --domains "example.com" --authenticator dns-cloudflare --dns-cloudflare-credentials "/etc/letsencrypt/credentials/credentials-3" --dns-cloudflare-propagation-seconds 240
Saving debug log to /var/log/letsencrypt/letsencrypt.log
Encountered CloudFlareAPIError adding TXT record: 10000 Authentication error
Error communicating with the Cloudflare API: Authentication error
Ask for help or search for solutions at https://community.letsencrypt.org. See the logfile /var/log/letsencrypt/letsencrypt.log or re-run Certbot with -v for more details.

    at ChildProcess.exithandler (node:child_process:397:12)
    at ChildProcess.emit (node:events:390:28)
    at maybeClose (node:internal/child_process:1064:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:301:5) 
```

although the API key is working fine 

```
curl -X GET "https://api.cloudflare.com/client/v4/user/tokens/verify" \
     -H "Authorization: Bearer xxxx" \
     -H "Content-Type:application/json"
{"result":{"id":"96ec8dc212843213fb16d363732e6b34","status":"active"},"success":true,"errors":[],"messages":[{"code":10000,"message":"This API Token is valid and active","type":null}]}
```

**Nginx Proxy Manager Version**
v2.9.14 
[...]

PR #2851, opened 2023-04-23
Fix for testing SSL certificates (Possibly Issue #2011)

S-T-3-V-3:
- Keepalive, User Agent + Accept headers
- Catch added for failed JSON parsing
- More accurate errors displayed to user

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -1181,9 +1181,8 @@ const internalCertificate = {
 						try {
 							const parsedBody = JSON.parse(responseBody + '');
 							resolve(parsedBody);
-						}
-						catch {
-							logger.warn("Error");
+						} catch (err) {
+    						logger.warn("Error: " + err.message);
 							logger.warn(`Status Code: ${res.statusCode}`);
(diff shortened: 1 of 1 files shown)


S-T-3-V-3:
This looks like Jenkins encountered an unrelated issue

backend/internal/certificate.js
```diff
@@ -1175,12 +1178,16 @@ const internalCertificate = {
 
 					res.on('data', (chunk) => responseBody = responseBody + chunk);
 					res.on('end', function () {
-						const parsedBody = JSON.parse(responseBody + '');
-						if (res.statusCode !== 200) {
+						try {
+							const parsedBody = JSON.parse(responseBody + '');
+							resolve(parsedBody);
+						}
+						catch {
```
jc21, line 1185: should be
```
} catch (err) {
    logger.warn("Error: " + err.message);
```

S-T-3-V-3:
Seems like some pretty inconsistent linting 🤔

jlesage:
Note that successfully parsing the body as JSON doesn't mean that the test was successful.  See https://github.com/NginxProxyManager/nginx-proxy-manager/pull/3388

PR #2898, opened 2023-05-08
Enable custom on location /

webysther:
With this change become easy to change location / without the need to put all HSTS and Websockets again inside advanced tab.


webysther:
Same #2816

PR #2899, opened 2023-05-08
Custom proxy by host

webysther:


Vaadasch:
@webysther 
Does the content of the file cannot simply be put inside the custom configuration inside the "Location" tab of the proxy_host ?
What is the difference between this field (the one displayed with the clog/gear icon) with your PR ?

webysther:
> @webysther Does the content of the file cannot simply be put inside the custom configuration inside the "Location" tab of the proxy_host ? What is the difference between this field (the one displayed with the clog/gear icon) with your PR ?

Automation, like the other options, the server don't have a well exposed API and with this is possible to use a decoupled which is not rewrite every time the host is saved.

PR #2929, opened 2023-05-18
Add strato.de to certbot dns plugins

FlixMa:
Add dns wildcard certificate support for strato.de using the provided certbot plugin.
See #1154 for previous discussion.

### DIFF
--- a/global/certbot-dns-plugins.js
+++ b/global/certbot-dns-plugins.js
@@ -521,6 +521,19 @@ aws_secret_access_key=wJalrXUtnFEMI/K7MDENG/bPxRfiCYEXAMPLEKEY`,
 		full_plugin_name: 'dns-route53',
 	},
 	//####################################################//
+	strato: {
+		display_name:        'Strato',
+		package_name:        'certbot-dns-strato',
+		version_requirement: '~=0.1.1',
+		dependencies:        '',
+		credentials:         `dns_strato_username = user
+dns_strato_password = pass
+# uncomment if domain name contains special characters
+# insert domain display name as seen on your account page here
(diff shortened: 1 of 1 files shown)


Substanzlos:
Tested the docker image, plugin works fine so far. Great work thanks!

Yoshi315161:
Hi guys,

i read this and wanted to try it out but i think i dont get it...
how and what for things do i have to put in the challange textfield?

also my strato is locked with 2fa
is there an example file i can use?

sorry but i dont understand the things under user and pass...
and for the SSL Domain i need "*.DOMAIN.COM"? or without the * for wildcard?

this is the example:
dns_strato_username = user
dns_strato_password = pass
# uncomment if domain name contains special characters
# insert domain display name as seen on your account page here
# dns_strato_domain_display_name = my-punicode-url.de

and now how to fill it?

the last two i dont understand...
do i have to remove the # and fill in something?
if i do i get an error.... 

i testet with:
*.DOMAIN.com

dns_strato_username = NUMBERS
dns_strato_password = PASSWORD
dns_strato_totp_secret = BUNCH OF NUMBERS AND CARACTERS
dns_strato_totp_devicename = NAME OF TOTP
# uncomment if domain name contains special characters <-- Leav this as it was
# DOMAIN.COM
# dns_strato_domain_display_name = *.DOMAIN.COM


pls help or point me a good example from someone who got it to work. it would be easier then the normal challenge then there are at time internal errors -,- (new request worked...)

thank you so much...

simonl169:
FYI

If your Strato PW contains a # as a character, you have to put the PW in quotation marks like this

dns_strato_password = "PASSWORD#"

PR #3000, opened 2023-06-15
Fix device or resource busy when patching IPv6 settings

xrh0905:
Caused by sed when mapping files into `/etc/nginx/conf.d`


jc21:
Thanks this was quite the annoyance!

PR #3049, opened 2023-07-09
Add bunny.net DNS to DNS challenges

deftdawg:
- Add support for bunny.net DNS challenges using @mwt's certbot-dns-bunny plugin.


deftdawg:
This is similar to #2790 but limited to only bunny.net stuff and done in a single commit to keep commit log tidy.

mwt:
Nice!

PR #3078, opened 2023-07-25
Corrected docker-compose.yml

andycandy-de:
The mysql folder should not be mounted to the npm container!


jc21:
Ah yes that was a mistake. Thanks!

Issue #3110, opened 2023-08-11
Sub path for proxied host


jerry-yuan:
**Is your feature request related to a problem? Please describe.**
I have deployed a NPM in docker to proxy the web services in other containers. Most of the backend services are work well entil I added a Guacamole, which was written in java with servlet.
The homepage of guacamole was `http://guacamole.host/guacamole`, whose suffix `guacamole` was unavoidable when using offical images. When I added a proxy host in NPM with domain `rdp.example.com`, I found I need to access the guacamole with `http://rdp.example.com/guacamole` and `http://rdp.example.com` displayed as a 404 page.
Although I can remove the suffix by repacking guacamole's image, I will loss the automatic update feature provided by watchtower which can update docker image from official repository.

**Describe the solution you'd like**
The nginx supports add a path on `proxy_pass`:
```
location / {
      proxy_pass ${proto}://${server}:${port}/${path}
      ... other configurations.
}
```
I'll appreciate if NPM can add a text box and support to append path suffix for `proxy_pass`.

**Describe alternatives you've considered**
I have tried to add some advanced custom configuration, but the `location /` is always overwrite by NPM and lead my custom configuration into a crash.

jerry-yuan:
keep issue open😌

PR #3115, opened 2023-08-12
add forward path prefix to prefix services in a sub path.

jerry-yuan:
I have opened an issue #3110  that I need to add some path prefix when forwarding requests, in order to proxy a sub dir into a independent domain.

I added a field on proxy host and proxy host's location configurations, which will be the path prefix of proxy_pass.

I am a Java backend developer. This is the first time I am modifying a project that uses Node.js as the backend implementation. If there are any inappropriate modifications, please feel free to correct my work.


jerry-yuan:
Is there any one can tell me what happend while running ci😭

jerry-yuan:
keep pr open

PR #3151, opened 2023-08-23
OCSP Stapling switch

BatSmacker84:
With this PR Dead, Proxy, and Redirect Hosts are now able to have OCSP Stapling enabled via the webgui with a simple toggle in the SSL tab.

The two config options added when OCSP Stapling is enabled are:
> ssl_stapling on;
> ssl_stapling_verify on;

The file used for OCSP Stapling is already provided by certbot (chain.pem), so all this PR does is allow for that stapling to occur by enabling those two options in each site's .conf file.
This is already possible in the current version of NPM by putting the two options into the Advanced Settings, but this PR allows for the config to look nicer and makes it easier for users to implement across their proxies.

The database does have to be migrated (a new migration file is included) in order to store if the Host has OCSP Stapling enabled or not.

### DIFF
--- a/backend/templates/_ocsp.conf
+++ b/backend/templates/_ocsp.conf
@@ -1,0 +1,7 @@
+{% if certificate and certificate_id > 0 -%}
+{% if ocsp_stapling == 1 or ocsp_stapling == true %}
+  # OCSP Stapling
+  ssl_stapling on;
+  ssl_stapling_verify on;
+{% endif %}
+{% endif %}
--- a/backend/doc/api.swagger.json
+++ b/backend/doc/api.swagger.json
@@ -96,6 +96,7 @@
 												"locations": [],
 												"hsts_enabled": 0,
 												"hsts_subdomains": 0,
+												"ocsp_stapling": 0,
 												"owner": {
 													"id": 1,
 													"created_on": "2023-03-30T01:11:50.000Z",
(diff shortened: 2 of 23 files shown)


backend/templates/_ocsp.conf
```diff
@@ -0,0 +1,7 @@
+{% if certificate and certificate_id > 0 -%}
+{% if ocsp_stapling == 1 or ocsp_stapling == true %}
+  # OCSP Stapling
+  ssl_stapling on;
+  ssl_stapling_verify on;
```
TheNytangel, line 5:
```suggestion
  ssl_stapling on;
  ssl_stapling_verify on;
{% if certificate.provider == "letsencrypt" %}
  ssl_trusted_certificate /etc/letsencrypt/live/npm-{{ certificate_id }}/chain.pem;
{% else %}
  ssl_trusted_certificate /data/custom_ssl/npm-{{ certificate_id }}/fullchain.pem;
{% endif %}
```
Should [ssl_trusted_certificate](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate) be included as well?

From the nginx docs:

> For verification to work, the certificate of the server certificate issuer, the root certificate, and all intermediate certificates should be configured as trusted using the [ssl_trusted_certificate](http://nginx.org/en/docs/http/ngx_http_ssl_module.html#ssl_trusted_certificate) directive.

Please note that, in my suggestion, I'm unsure about the custom SSL chain 🙂

Issue #3138, opened 2023-08-16
DNS challenge with DuckDNS does not work


Rihan9:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**

This isseus is some sort of "reopening" of #2092. We can't request a SSL certificate for a domain hosted on DuckDNS.
In the log we found the following message:
`"The DNS response does not contain an answer to the question: mydomain.duckdns.org. IN TXT"`

**Nginx Proxy Manager Version**

v2.10.3

**To Reproduce**
1. create a hostname on duckDNS and make sure is pointed to the nginx proxy manager installation
2. on tab ssl certificates --> add ssl certificates --> let's encrypt --> compelte the form using dns challange: yes and the required token (see screenshot)
3. the error should pop-up

**Screenshots**

![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/33684020/785ec7cb-5c61-4178-b10a-96c91f0aa684)
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/33684020/e4a95ae7-e413-4827-82a2-1f107560b11d)
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/33684020/d0f0ce78-d883-4ae7-b788-ddd584ec5e12)

holocronology:
I had the same issue earlier today when testing something with duckdns.org. Now, because the duckdns challenge failed, my entire NPM is borked.

![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/65265114/e3d67a50-4f31-4ebb-bb6d-fe1f0d167717)

holocronology:
Did you delete a cert prior to this? I think there is an error in NPM that the database is not updating when certs are deleted. So NPM starting looking for the folder of the deleted cert, but it is not there. SO it ends up in an error loop looking for these deleted files.

Rihan9:
I don't think so. The only think that come to my mind is the renew certifica process. Can it cause the deletion of the old certificate?

devedse:
I'm running into exactly the same issue. I've tried to clean everything up an recreate the certificate but no matter what I tried I keep having this issue.

devedse:
Ok I hacked the solution to work.

[...]

PR #3155, opened 2023-08-24
Added force renewal + --dns-duckdns-no-txt-restore

devedse:
This should fix:
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3138

@jc21 up to you if this is the right way to fix it.

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -908,6 +908,10 @@ const internalCertificate = {
 			mainCmd = 'AWS_CONFIG_FILE=\'' + credentialsLocation + '\' ' + mainCmd;
 		}
 
+		if (certificate.meta.dns_provider === 'duckdns') {
+			mainCmd = mainCmd + ' --dns-duckdns-no-txt-restore';
+		}
+
 		logger.info('Command:', `${credentialsCmd} && ${prepareCmd} && ${mainCmd}`);
 
 		return utils.exec(credentialsCmd)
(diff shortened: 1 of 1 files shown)


jc21:
Seems fine to me. Let me know if this image works for you.

devedse:
I'll give it a try. Do you know enough about certbot to be able to explain why my change is needed 😄 ?

jc21:
Well, no I don't know why _not_ restoring the original TXT record would be the answer to this problem. I fully understand the DNS verifiction process, but I feel that the devs of the duckdns certbot might have a bug there.

devedse:
I just tried it with the normal unchanged code (:latest image) and it works again... wtf... My guess is that DuckDNS can be in a specific corrupted state or something where we need need to not restore the OriginalTxtRecord???

It might still be good to merge this since it apparently can fix the issue.

g1za:
Hello,
I'm now getting the following error:
`CommandError: usage: 
  certbot [SUBCOMMAND] [options] [-d DOMAIN] [-d DOMAIN] ...

Certbot can obtain and install HTTPS/TLS/SSL certificates.  By default,
it will attempt to use a webserver both for obtaining and installing the
certificate. 
certbot: error: unrecognized arguments: --dns-duckdns-credentials /etc/letsencrypt/credentials/credentials-113 --dns-duckdns-no-txt-restore

    at /app/lib/utils.js:16:13
    at ChildProcess.exithandler (node:child_process:410:5)
    at ChildProcess.emit (node:events:513:28)
    at maybeClose (node:internal/child_process:1100:16)
    at Process.ChildProcess._handle.onexit (node:internal/child_process:304:5)`
    
Did they maybe changed anything on DuckDNS side and `--dns-duckdns-no-txt-restore` is not needed anymore?
Let me know if you think it is the right way and I can create a new PR (sorry but I'm not sure I can test on my machine).
Thanks!

PR #3247, opened 2023-10-08
Host should be $proxy_host, not $host

### LINKED ISSUE #2675: Cannot proxy to sites that require a correct Host header
This line...

https://github.com/NginxProxyManager/nginx-proxy-manager/blame/develop/docker/rootfs/etc/nginx/conf.d/include/proxy.conf#L2

I believe should be...

`proxy_set_header Host $proxy_host;`

(I got the below to work by doing it)

Without that, proxy requests are being forwarded to websites with the original Host header and those sites are returning an error or a 404. For example, trying proxying data.example.com to an S3 bucket like mybucket.s3-website-us-east-1.amazonaws.com:80

S3 will reply with something like..

```
404 Not Found
Code: NoSuchBucket
Message: The specified bucket does not exist
BucketName: data.example.com
RequestId: J8K24Y80A04JNFY4
HostId: ksf8GNLOX46prb58795rY6bxvr81Lfm6xzKnWTJcXLEaomW69xYWmBJBvcWV4VMZJEZohq4CF4s=
```

Further, you cannot override this proxy_set_header command by adding it to the advanced custom config in the UI because it simply appends the value to the existing header. ie, S3 will reply as blow (note the BucketName)...

```
404 Not Found
Code: NoSuchBucket
Message: The specified bucket does not exist
BucketName: data.example.com mybucket.s3-website-us-east-1.amazonaws.com
```

Nor have I found a way to "unset" proxy_set_header Host. I tried using "Headers More" to no avail: https://www.nginx.com/resources/wiki/modules/headers_more/

vbraun:
* $host is the Host sent by the user's browser. This is the correct setting if the service has built-in support for being proxied.

* $proxy_host is the Host as if the browser would run on the proxy. This is the correct setting if the service does not have built-in support for reverse proxies. It is also the nginx default.

* In nginx, you cannot unset the Host header. Configuring headers multiple times just sends multiple values with the http request. So there is no way to "fix" the Host by adding a custom header if it is already set.

For these reasons, Host should not be set (and default to $proxy_host). In the unlikely case that your service needs something else you can then just set a custom Host header in the GUI.

Fixes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/2675

### DIFF
--- a/docker/rootfs/etc/nginx/conf.d/include/proxy.conf
+++ b/docker/rootfs/etc/nginx/conf.d/include/proxy.conf
@@ -1,5 +1,4 @@
 add_header       X-Served-By $host;
-proxy_set_header Host $host;
 proxy_set_header X-Forwarded-Scheme $scheme;
 proxy_set_header X-Forwarded-Proto  $scheme;
 proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;


valinet:
How is this STILL not merged? Such a design flaw just obliterates all the usefulness this program brought. PLEASE MERGE THIS and let's have it fixed at last.

kvchoi:
proxy_set_header Host $host;

KevinMitchell:
> more_set_input_headers 'Host: [www.domain-new.com](http://www.domain-new.com)';
...
> so, you can reset the input header by command : more_set_input_headers

Not sure I understand how more_set_input_headers helps here.  It looks like any custom headers I write, including more_set_input_headers, will appear in the generated conf file before the generated `proxy_set_header Host $host;` line.  So are you saying more_set_input_headers takes precedence over any proxy_set_header statements, even if it appears before such statements?

PR #3258, opened 2023-10-15
Removed /etc/letsencrypt from explicit volumes

### LINKED ISSUE #3170: Ways to move letsencrypt storage to /data volume
Hello, I create addon for Home Assistant OS which are just Docker image with some integration around.

For it I need to have all persistent data stored only in /data, other volumes will not survive recreate.

But there is an issue with /etc/letsencrypt - it is volume, so I cannot create symlink for it in my image (using your image as base).

I would like to remove /etc/letsencrypt from VOLUME in the Dockerfile.
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/fa851b61da3fe3726d1a04c25e69d36e79edea2d/docker/Dockerfile#L54

Or have the directory configurable, but this seems more work.

There is option to not use this image as a base, but copy&paste its content to my Dockerfile, but this will be maintenance hell.

Thank you for help. I can do PR if you confirm the solution.

iBobik:
So it can be moved in other images using this as a base.

Fixes #3170

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -51,7 +51,7 @@ RUN rm -rf /etc/s6-overlay/s6-rc.d/user/contents.d/frontend /etc/nginx/conf.d/de
 	&& pip uninstall --yes setuptools \
 	&& pip install --no-cache-dir "setuptools==58.0.0"
 
-VOLUME [ "/data", "/etc/letsencrypt" ]
+VOLUME [ "/data" ]
 ENTRYPOINT [ "/init" ]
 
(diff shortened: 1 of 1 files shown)


jc21:
CI is failing because the lets encrypt folder needs to be mounted and there's a check to enforce that.

Modify [this file](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/docker/docker-compose.ci.yml) as part of your patch and add a named volume to the `fullstack-*` services, similar to `npm_data` volume.

Issue #853, opened 2021-01-31
SSL PassThru


ctops1:
**Is your feature request related to a problem? Please describe.**
Unable to pass through encrypted SSL connections to a server that doesn't need SSL encryption

**Describe the solution you'd like**
A clear option or configuration to enable SSL passthrough via stream or similar method. 

**Describe alternatives you've considered**
Attempted to add code from links under additional context, host simply goes offline.

**Additional context**
This has been discussed in some other unclosed issues linked at the bottom, but this should be possible with modern NGINX using the stream function. I see that NPM has "Stream" but it only allows you to capture an entire port instead of a specified domain for forwarding to another internal host. 

https://serversforhackers.com/c/tcp-load-balancing-with-nginx-ssl-pass-thru
https://stackoverflow.com/questions/38371840/ssl-pass-through-in-nginx-reverse-proxy

https://github.com/jc21/nginx-proxy-manager/issues/235
https://github.com/jc21/nginx-proxy-manager/issues/461

ma-karai:
I have to say this would be a big feature. This would make dealing with cooler larger projects a breeze. Matrix/synapse comes to mind which is a pain to setup behind a separate proxy. It would be nice if you could just forward a specific domain 80 and 443 traffic to a service on a separate machine that does the ssl stuff.

ctops1:
After spending a lot of time getting this to work in NGINX itself - I realize this request may not fulfill everyone's needs even if it was implemented according to what NGINX supports.

NGINX can do SSL passthru, but it can't do it on the same ports being serviced for normal NGINX services. You can do only one or the other. 

For example, let's say you have some sites on 443 using normal NGINX SSL. You cannot then also use NGINX to perform passthrough to a VM that has its own certificates. You would have to use a different port, such as 8443 or a custom port. The only other alternative would be another box and another public IP.

I finally resolved my issue by switching back to pure NGINX and implementing passthru on 8443 and the rest of our sites on 443.

NPM is a great tool but it has its limitations - perhaps someday I will revisit it and I will still recommend it to those who have less complex setups than I.

centralhardware:
for ssl pasthrough you need to use stream (npm now support it), but i think you can't use stream and http on the same port.

jdkruzr:
> I was looking for this feature as well. Since the above PR #1479 doesn't seem to have been merged, I baked my own quick fix. It's not pretty but seems to do the trick (not tested thoroughly yet). Sharing here as it might help others.
> 
[...]

PR #1479, opened 2021-10-12
SSL passthrough hosts

chaptergy:
This PR would resolve https://github.com/jc21/nginx-proxy-manager/issues/853.

SSL passthrough would be a new type of host where the ssl certificate of the upstream server is used, so no ssl termination is done at the proxy. But the only way this is possible in nginx is streams, using SNI to forward the packet to the right destination. As a stream and the normal proxies using http cannot be on the same port, **all** https traffic has to go through the stream, which then could forward it internally to the http proxy. As this reduces performance for all hosts, this feature is strictly opt-in. The documentation was updated to add a section about this to the advanced config page.

PR #3331, opened 2023-11-17
Ssl passthrough hosts - updated

acul009:
Hi,

This PR updates the original Request:  [SSL passthrough hosts #1479](https://github.com/NginxProxyManager/nginx-proxy-manager/pull/1479)

My use case is passing data to hosts which can handle certificate request for themselves.
In those cases I'd use the http proxy to only forward the requests to /.well-known/acme-challenge.

Additional traffic then has to arrive over TLS to work.
This makes reusing Port 443 for multiple services a breeze.

@chaptergy If you have some time, I'd really appreciate if you could check that I didn't botch any of your work :)

This would resolve Ticket #853 and allow for more secure data handling inside local networks.


cj0duke:
> Docker Image for build 2 is available on [DockerHub](https://cloud.docker.com/repository/docker/jc21/nginx-proxy-manager) as `jc21/nginx-proxy-manager:github-pr-3331`
> 
> **Note:** ensure you backup your NPM instance before testing this PR image! Especially if this PR contains database changes.

I'd like to test the feature. I did pull the image 3331, the stack is starting without problems, but within NPM - SSL P Hosts there is a message "SSL Passthrough Hosts are not enabled in the environment. Please see the docs for more information." Cloud you please provide some information on how to enable SSL P in the environment?

acul009:
Right, the documentation is only contained in the branch, so you wouldn't see thatnormally.

The corresponding Environment Variable is:
ENABLE_SSL_PASSTHROUGH

Just set that to true and the new option can be found in the webinterface.

cj0duke:
> Just set that to true and the new option can be found in the webinterface.

Thank you for the quick update. It works.

Unfortunately when accessing a passthrough host, firefox is returning "SSL_ERROR_UNRECOGNIZED_NAME_ALERT".

(Maybe I need a coffee or two)

acul009:
Can you describe your configuration a bit?

You might already know this, but I'll describe how this works a bit so we're on the same page.
Its hard to know how much the person on the other end of the line known on the internet ;)


This is only a SSL/TLS passthrough which uses the SNI from the TLS protocoll to route the raw traefik.
Your error sounds like a problem with the hosts certificate.
It's important, that the connection with your given address (test.example.com) would work if you just route the whole port through.

So you can't just add test2.example.com as a passthrough if you target system doesnt have the required certificate.


If that helped, great.
In case you already know all that you'll have to share a bit more information to find the cause :)

cj0duke:
> If that helped, great. In case you already know all that you'll have to share a bit more information to find the cause :)

I tried it once again, with fresh docker images, but same result.

Setup as follows:
vm with working ssl cert serving with apache
bind running (seperate) to serve domain

Did the following steps:
- bind entry pointing to vm
- changed entry to point to npm
- setup npm for passthrough (see pic)
npm

Still getting the firefox error of "SSL_ERROR_UNRECOGNIZED_NAME_ALERT"

acul009:
unfortunately it seems like this won't be merged anytime soon, even if I updated it.

In case anyone direly needs a solution for TLS forwarding, I created another project (WIP):

https://github.com/Rahn-IT/traefik-gui.git

Issue #1005, opened 2021-04-09
Unnecessary HSTS header over HTTP


DarioViva42:
![grafik](https://user-images.githubusercontent.com/45972949/114214087-2adf7900-9964-11eb-8d9e-5851c6dda60e.png)

I wanted to add my Website to the hsts-preload list: https://hstspreload.org/
With the current configuration I was able to add my site to the list. But I still have this very annoying message.
The hsts header should only be added if i access the page with https, but not when i access it over http.
![grafik](https://user-images.githubusercontent.com/45972949/114216081-b8bc6380-9966-11eb-9810-c325afdf28b5.png)

I have tried a lot of different methods over the last days, but none seemed to work. First I tried to add the header in my node application instead of in the npm (not node package manager, but nginx proxy manager). But this obviously does not work, as the https terminates at the npm. so req.connection.enrypted is always undefined.

```js
var proto = req.connection.encrypted ? 'https' : 'http';
if (proto =="https") {
  res.setHeader("Strict-Transport-Security", "max-age=31536000;includeSubDomains; preload");
}
```

I also tried to solve it in the advanced settings:
![grafik](https://user-images.githubusercontent.com/45972949/114215573-097f8c80-9966-11eb-82c8-b148740be623.png)

but this immediately makes the applikation stop working.
![grafik](https://user-images.githubusercontent.com/45972949/114215749-464b8380-9966-11eb-9267-a5cba4f2675a.png)

It really seemes that I am not able to solve this problem on my own and I need your help. I think the problem needs to be tackled inside npm itself.

Kind Regards

Dario Viva

l4rm4nd:
Put the following line into your advanced configuration of npm:

`include conf.d/include/force-ssl.conf;`

![image](https://user-images.githubusercontent.com/21357789/122946278-87561000-d379-11eb-9c9a-247f9595ef6f.png)

This ensures that your site is only accessible by an encrypted communication channel (HTTPS). If a request occurs on HTTP, a redirect to HTTPS will take place.

DarioViva42:
@l4rm4nd sorry for taking so long to respond to you (wow its more than two years past)
I just wanted you to inform you that your solution sadly did not work.
But I found something that should probably work.
https://websistent.com/add-the-hsts-header-only-for-https-requests-nginx/
but this configuration can probably not changed easily by myself.
when i add map into  the Custom Nginx Configuration field the service goes offline.

DarioViva42:
[...]

PR #3360, opened 2023-12-02
only add hsts header with https.

DarioViva42:
fixes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1005 for more information look at: https://websistent.com/add-the-hsts-header-only-for-https-requests-nginx/

It would be really great if you could check if this solution really works.


DarioViva42:
well that’s a bummer, I can not see what failed, because obviously I don’t have an account for your Jenkins server.

Issue #3218, opened 2023-09-26
data.domain_names should NOT have more than 15 items << Multi Domain SAN Certs Limit


abdulfahad661:
**Is your feature request related to a problem? Please describe.**
Its not a problem looks like but a limit that should be increased i suggest, we use Route 53 and Cloud Flare and we have normal request of SAN for 30 to 40 sites at a time and looks there is limit set by developers in this NPM which can be great if increased.

<img width="194" alt="15 Items limits in domian names" src="https://github.com/NginxProxyManager/nginx-proxy-manager/assets/136641221/7407edfd-1d31-4dbb-b4ee-f898a3cdbfbd">

**Describe the solution you'd like**
Increase limit to like upto 50 for SAN sites (multiple domains at once)
Looks its been controlled at NginxProxyManager/nginx-proxy-manager/blob/develop/frontend/js/app/nginx/certificates/form.js
Line 268 >>             maxOptions:   15,
If that can be increased so that can be increased as i guess (i have no dev experience)

<img width="438" alt="max options" src="https://github.com/NginxProxyManager/nginx-proxy-manager/assets/136641221/74a69fd8-a756-4d0e-92fb-56acdae45fc7">

**Describe alternatives you've considered**
We do this by right now on linux using command below and it works for more that 15 domains(we have to enter DNS challenge for each item) :
certbot-2 certonly --email email@email.com  -d agent.ha.domain.com -d auth.ha.domain.com -d billing.ha.domain.com -d bots.ha.domain.com -d cm.ha.domain.com -d docs.ha.domain.com -d eps.ha.domain.com -d onboarding.ha.domain.com -d recordings.ha.domain.com -d reports.ha.domain.com -d security.ha.domain.com -d smgr.ha.domain.com -d wfm.ha.domain.com -d wizard.ha.domain.com -d www.ha.domain.com --agree-tos --manual --preferred-challenges dns
 

**Additional context**
This NPM is great and this limit increase can make life more easy :)

Shocktrooper:
I just ran into this.  Unless there is a limitation on nginx or lets encrypt this field should reflect how many you can actually fit in a cert or nginx

abdulfahad661:
Yes I hope that NPM devs can assist on enhancing this limits

Shocktrooper:
If someone can point me to the code that needs to be changed I could handle this myself, or someone else if they get to it first

abdulfahad661:
[...]

PR #3382, opened 2023-12-12
fix: increasing maxOptions (amount of domains) to 30

r3na:
ref:
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3218

### DIFF
--- a/backend/schema/definitions.json
+++ b/backend/schema/definitions.json
@@ -172,7 +172,7 @@
       "description": "Domain Names separated by a comma",
       "example": "*.jc21.com,blog.jc21.com",
       "type": "array",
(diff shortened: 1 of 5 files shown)


r3na:
@jc21  I tested locally with more than 15 domains , also generating SSL certificates, it seems to be fine. if you could take a look, this would be very helpful to be included with the next release :)

Shocktrooper:
As a suggestion we should probably make this inline with what LetsEncrypt allows.  Shouldn't be a big change
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3218#issuecomment-1851527067

Issue #1317, opened 2021-08-13
Make logrotate configurable


Apollon77:
**Is your feature request related to a problem? Please describe.**
I use nginx-proxy-manage rnow since 3 days and one logfile is already 6GB ... This is way too much.

I found https://github.com/jc21/nginx-proxy-manager/pull/1140, but 1 weeks is simply looong ... I do not need that. a standard daily rotate for 7 days or such is completely sufficient for my usecase

**Describe the solution you'd like**
Allow lofgrotate settings to be configurable, at least the interval and generations

**Describe alternatives you've considered**
I thought about https://github.com/jc21/nginx-proxy-manager/issues/183#issuecomment-534536613 but do not know if/how it interfers with the now integrated logrotate

I will for now increase storage, but I would like to see that only a temporary solution ...

chaptergy:
If you want you can integrate new environment variables which generate a logrotate config on boot and create a PR for this.

You could also just mount a custom logrotate config into the container at `/etc/logrotate.d/nginx-proxy-manager` via your docker-compose. For reference, [here is the current config](https://github.com/jc21/nginx-proxy-manager/blob/develop/docker/rootfs/etc/logrotate.d/nginx-proxy-manager).  
However the logrotate binary is only called [every two days](https://github.com/jc21/nginx-proxy-manager/blob/b0dc68d7d487d628c825efd3718e4cacbc49ff36/backend/setup.js#L210), so that is the smalles possible rotation period possible without changing the source code.

Apollon77:
@chaptergy I'm not experienced enough on docker side to make a PR and also for the other "mount custom config" I would need some more guidance ... sorry but in this regard I'm complete noob :-( ... every two days could work in general I think ... could you please provide an example on how to get that in via docker-compose?

chaptergy:
Create a file on the machine running docker with your custom logrotate config. Then just add the following line to your docker-compose:

```yml
# version: "3"
# services:
#   app:
#     ...
#     volumes:
        - <path-to-your-config-file>:/etc/logrotate.d/nginx-proxy-manager
#     ...
```

The commented out lines are only there for reference so you know where to put it, the only relevant line to add is the uncommented one.

Apollon77:
Ok, tried that for now, lets see how it goes. I will see in 2 days ... But I think allowing that to be configured would be a good idea or?!

chaptergy:
[...]

PR #3422, opened 2023-12-30
Add documentation on customising logrotate config

Encephala:
As discussed in #1317, it isn't feasible to add configuration options to the compose file for customising logrotate functionality, as way too many options exist.

However, customising the config isn't very hard to do, so added a section to the docs on the Advanced Configuration page.


Encephala:
Whoops sorry for the force push, I pushed to the wrong branch

jgkennedy:
`./logrotate.custom:/etc/logrotate.d/nginx-proxy/manager` should be `./logrotate.custom:/etc/logrotate.d/nginx-proxy-manager` (last slash should be a hyphen)

Issue #3447, opened 2024-01-12
websupportsk plugin not working correctly, API update needed?


lkxnon:
**What provider would you like to see added to NPM?**
Websupport.sk

**Have you checked if a certbot plugin exists?**
There is newer version of  DNS Authenticator plugin for Certbot
https://pypi.org/project/certbot-dns-websupport/ (version certbot-dns-websupport 2.0.1).
Seems it is fork from version 0.1.6 used in NPM. I have tested this version 2.0.1 and it works fine from Docker.

This one seems to be faulty and it is quite outaded (2years). There are 404 errors when using this plugin.
```
websupportsk: {
		display_name:        'Websupport.sk',
		package_name:        'certbot-dns-websupportsk',
		version_requirement: '~=0.1.6',
		dependencies:        '',
		credentials:         `dns_websupportsk_api_key = <api_key>
dns_websupportsk_secret = <secret>
dns_websupportsk_domain = example.com`,
		full_plugin_name: 'dns-websupportsk',
```

This is error in logs from NPM when trying to use Websupport.sk plugin:

```
2024-01-10 22:04:19,330:DEBUG:acme.client:Storing nonce: xavOoKHvrBw7alE4Os1fNMF21tAV-IcbMRof84ecu6iECyKuT_8
2024-01-10 22:04:19,332:INFO:certbot._internal.auth_handler:Performing the following challenges:
2024-01-10 22:04:19,332:INFO:certbot._internal.auth_handler:dns-01 challenge for hass.kostrab.sk
2024-01-10 22:04:19,344:DEBUG:urllib3.connectionpool:Starting new HTTPS connection (1): rest.websupport.sk:443
2024-01-10 22:04:19,686:DEBUG:urllib3.connectionpool:[https://rest.websupport.sk:443](https://rest.websupport.sk/) "GET /v1/user/self HTTP/1.1" 200 None
2024-01-10 22:04:19,805:DEBUG:urllib3.connectionpool:[https://rest.websupport.sk:443](https://rest.websupport.sk/) "POST /v1/user/self/zone/hass.kostrab.sk/record HTTP/1.1" 404 None
2024-01-10 22:04:19,816:DEBUG:certbot._internal.display.obj:Notifying user: Waiting 120 seconds for DNS changes to propagate

.
.
.
.
.
.
2024-01-10 22:06:24,424:DEBUG:certbot._internal.error_handler:Calling registered functions
[...]

PR #3513, opened 2024-01-30
Updated certbot-dns-websupport plugin to 2.0.1 #3447

setrin:


setrin:
Tested, works without problem. Fixes #3447
Please merge

PR #1882, opened 2022-02-23
PROXY Protocol support

SBado:
This PR add basic support for [PROXY protocol](https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol) to NPM. I needed NPM to support PROXY protocol becasue I'm using two instances of HAProxy as a point of access to my services, as described [here](https://theorangeone.net/posts/server-2020-proxmox/#ingress). It's been working fine for me so far, so I thought to share.

Note: to allow coexistence of "regular" and "PROXY protocol enabled" hosts, the latter ones will listen on port 88 and 444.

Related issue: #1114.

immagine

PR #3537, opened 2024-02-09
PROXY Protocol support

jwklijnsma:
https://github.com/NginxProxyManager/nginx-proxy-manager/pull/1882 update code base


jwklijnsma:
@jc21  where can i fine the output of the failed test ?

Morriz:
bump

jc21:
```
CypressError: `cy.task('backendApiPost')` failed with the following error:

> 400: tag "endif" not found, file:/app/templates/_listen.conf, line:20, col:1, line:12, col:1
```

jwklijnsma:
> ```
> CypressError: `cy.task('backendApiPost')` failed with the following error:
> 
> > 400: tag "endif" not found, file:/app/templates/_listen.conf, line:20, col:1, line:12, col:1
> ```
What for test is this in the background ?

jc21:
```
Integration Tests / Mysql / Hosts endpoints.Should be able to create a http host

Error Message:
data should NOT have additional properties
```

There is a swagger/openapi schema file that acts as a contract for the API, can be [found here](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/backend/doc/api.swagger.json). This isn't generated from code on purpose, as it's meant to prevent people making breaking changes by accident.

To get past this, since you're adding new fields to the responses of endpoints, you'll need to define them in the swagger file as well.

PR #3581, opened 2024-02-23
Fix: increase max number of domains to 100 (match with Let's Encrypt)

davidindra:
CC @jc21 

Let's make the limit of domains count equal to the Let's Encrypt limit, as suggested here https://github.com/NginxProxyManager/nginx-proxy-manager/pull/3382#issuecomment-1854005778, when the limit was raised from 15 to 30 in last December.

In my use case, I have a big SAN cert - lots of subdomains and "www." is needed to be supported too, so everything twice; I am not able to configure this in nginx-proxy-manager at the moment.

ref: https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3218, https://github.com/NginxProxyManager/nginx-proxy-manager/pull/3382

I would be glad if you could approve & merge this change :)
Thx and have a nice day,

D.

### DIFF
--- a/backend/schema/definitions.json
+++ b/backend/schema/definitions.json
@@ -172,7 +172,7 @@
       "description": "Domain Names separated by a comma",
       "example": "*.jc21.com,blog.jc21.com",
       "type": "array",
-      "maxItems": 30,
+      "maxItems": 100,
       "uniqueItems": true,
       "items": {
         "type": "string",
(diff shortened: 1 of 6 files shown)


davidindra:
Thanks @jc21! 🙂 Could you say when you will make a new release to the :latest Docker tag?

PR #3589, opened 2024-02-27
use client provided header to determine access protocol

### LINKED ISSUE #3365: force ssl with cloudflared not working
the toggle is enabled, and the inclusion of the force-ssl.conf file is part of the proxy config; however, curl -v http://some.example.com returns without a redirect to https

if relevant, npm is exposed via cloudflared with config like:
```
tunnel: ***
credentials-file: ***

ingress:
  - service: https://nginx-proxy-manager
    originRequest:
      originServerName: proxy.example.com
```

realies:
this helps when using tunnels like cloudflared and fixes #3365


jc21:
I believe that this approach will only work when NPM is behind another proxy, like Cloudflare as per the issue. However, might have unexpected problems when hitting directly.

PR #3600, opened 2024-03-03
Update certbot-dns-duckdns version (fix #2994)

### LINKED ISSUE #2994: Can't use both Google Domains DNS and DuckDNS for SSL Certificates
**Checklist**
  - Yes
  - Yes
  - Yes

**Describe the bug**
tl;dr You can't install both certbot-dns-google-domains 0.1.11 and certbot-dns-duckdns 0.9 at the same time.

You can't use both GoogleDomainsDNS and DuckDNS at the same time. I originally set up certificates using duckdns, but then wanted to set up a personal personal domain so I purchased a domain using Google domains. Unfortunately you can't use both at the same time. certbot-dns-duckdns depends on zope.interface 5.4 and certbot-dns-google-domains depends on zope.interface 6.0 and they're incompatible. The error from pip looks like this:

Error: pip's dependency resolver does not currently take into account all the packages that are installed. This behavior is the source of the following dependency conflicts.
certbot-dns-duckdns 0.9 requires zope.interface~=5.4.0, but you have zope-interface 6.0 which is incompatible.

2.10.3

**To Reproduce**
1. Go to SSL Certificates
2. Click on Add
3. Add a ssl certificate for DuckDNS
4. Afterward, try to add another certificate for GoogleDomainsDNS

**Expected behavior**
You should be able to add SSL certificates for both domain providers

**Screenshots**

**Operating System**
This was hosted on Debian, but should not be dependent on host OS

**Additional context**
N/A

sdaqo:
So this fixes #2994, a version conflict that can be solved by updating the duckdns certbot version

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -90,7 +90,7 @@
 	"duckdns": {
 		"name": "DuckDNS",
 		"package_name": "certbot-dns-duckdns",
-		"version": "~=0.9",
+		"version": "~=1.0",
 		"dependencies": "",
 		"credentials": "dns_duckdns_token=your-duckdns-token",
 		"full_plugin_name": "dns-duckdns"


jc21:
@sdaqo can you confirm the provider is working with the docker image above?

sdaqo:
Yes I am running this fork on my own because I am running into the problem myself and I can confirm that it works for me.

PR #3642, opened 2024-03-20
fix: update certbot-dns-transip to latest version

leinelissen:
This updates the `certbot-dns-transip` package to the latest version, since the current version is incompatible with certbot 2.0.

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -418,7 +418,7 @@
 	"transip": {
 		"name": "TransIP",
 		"package_name": "certbot-dns-transip",
-		"version": "~=0.4.3",
+		"version": "~=0.5.2",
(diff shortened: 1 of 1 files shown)


wokkeltje13:
When will this be released? I need a cert renew in 3 weeks that is currently not working.

leinelissen:
It will most probably take more than three weeks. You can either use the the image build for this PR (we're running it and we're not running into any issues at the moment), or you can upgrade the package manually in your container.

wokkeltje13:
I have modified /app/global/certbot-dns-plugins.json but nothing changed. how can i update the plugin in my environment?

leinelissen:
I think this did it for us:
```sh
pip install certbot-dns-transip ==0.5.2
```

PR #3654, opened 2024-03-26
Map volumes to persistent paths in example compose files

he3als:
The documentation currently includes example `docker-compose.yml` files that will clear users' NPM data when the server is re-deployed, which can confuse, e.g., [this Reddit post](https://www.reddit.com/r/nginxproxymanager/comments/v8i7ju/nginx_proxymanager_data_removing_always_aftest/).

This pull request makes the example compose files instead save to a persistent location, `/srv/nginx-proxy-manager`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ services:
       - '81:81'
       - '443:443'
     volumes:
-      - ./data:/data
-      - ./letsencrypt:/etc/letsencrypt
+      - /srv/nginx-proxy-manager/data:/data
+      - /srv/nginx-proxy-manager/letsencrypt:/etc/letsencrypt
 ```
 
 This is the bare minimum configuration required. See the [documentation](https://nginxproxymanager.com/setup/) for more.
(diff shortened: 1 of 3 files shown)


jc21:
I disagree with this change. The docker-compose example will create persistent, mounted volumes relative to the CWD when the `compose up` is run. The only way these volumes are wiped out is when running `compose down` or if the CWD is itself a temporary FS.

The reddit post does not determine _why_ mounting an absolute folder worked for them.

he3als:
> I disagree with this change. The docker-compose example will create persistent, mounted volumes relative to the CWD when the `compose up` is run. The only way these volumes are wiped out is when running `compose down` or if the CWD is itself a temporary FS.
> 
> 
> 
> The reddit post does not determine _why_ mounting an absolute folder worked for them.

It's from running `compose down` or recreating the containers.

Many people recreate the containers as a stack in Portainer to change settings, which unexpectedly removes all data.

However, maybe it would be best to move to a persistent volume instead of an absolute folder?

PR #3691, opened 2024-04-09
Update certbot-dns-infomaniak

Fuechslein:
Latest plugin version with several fixes.


jc21:
@Fuechslein can you confirm the provider is working with the docker image above?

Issue #3580, opened 2024-02-23
] [Nginx    ] › ⬤  debug     Could not delete file: { app-1  |  


Jordibgzashtita:
HI.

I try add 204 response to domain.com and it go es offline. So i delete it from data.... folder and same way it was, and again it go to red.... Include i delete everything from scratch and same app-1  | [2/23/2024] [1:53:39 PM] [Nginx    ] › ⬤  debug     Could not delete file: {
app-1  |   "errno": -2,
app-1  |   "code": "ENOENT",
app-1  |   "syscall": "unlink",
app-1  |   "path": "/data/nginx/proxy_host/1.conf"
app-1  | }

This is from clean config.... 

When try add custom location it goes offline

Jordibgzashtita:
![Uploading IMG_20240223_133500.jpg…]()

Jordibgzashtita:
this config is created from nginxproxymanager github-pr-3570 i returned to old docker and it work

Klausberger:
Same error when changing or adding proxy host in the webfrontend:

![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/42064202/5e071e00-f1b5-4e5d-a9ca-0941f3d16d39)

Klausberger:
> this config is created from nginxproxymanager github-pr-3570 i returned to old docker and it work

Hi, what docker-image version is that, that is working for you?

matteopasini:
> Same error when changing or adding proxy host in the webfrontend:
> 
[...]

Issue #677, opened 2020-10-26
Customize SSL cert renewing interval to avoid reloading nginx every hour


niklasdahlheimer:
**Is your feature request related to a problem? Please describe.**
First of all: great tool! Setting up reverse poxies is fun with this one :)  

According to the logs, npm is checking for expired certificates every hour:
```
[10/26/2020] [1:26:22 AM] [SSL      ] › ℹ  info      Renewing SSL certs close to expiry...
[10/26/2020] [1:26:24 AM] [Nginx    ] › ℹ  info      Reloading Nginx
[10/26/2020] [1:26:24 AM] [SSL      ] › ℹ  info      Renew Complete
[10/26/2020] [2:26:22 AM] [SSL      ] › ℹ  info      Renewing SSL certs close to expiry...
[10/26/2020] [2:26:24 AM] [Nginx    ] › ℹ  info      Reloading Nginx
[10/26/2020] [2:26:24 AM] [SSL      ] › ℹ  info      Renew Complete
[10/26/2020] [3:26:22 AM] [SSL      ] › ℹ  info      Renewing SSL certs close to expiry...
[10/26/2020] [3:26:23 AM] [Nginx    ] › ℹ  info      Reloading Nginx
[10/26/2020] [3:26:23 AM] [SSL      ] › ℹ  info      Renew Complete
[...]
```
I think reloading nginx every hour could lead to problems in the business times of our web services (e.g. downloads buffered by nginx ..)

**Is there a way to customize the cron job which triggers the cert renewing every hour to start the renewing only once a day or so?** 

**Describe the solution you'd like**
some way (UI or production.config) to customize the renawal interval

tophee:
I don't quite understand why the renewing interval is set to such a short period in the first place. I think, letsencrypt certs are usually valid for three months or so. Why do 2160 checks during that period?

niklasdahlheimer:
probably it's just changing
`intervalTimeout:    1000 * 60 * 60, // 1 hour`
to
`intervalTimeout:    (process.env.CERTIFICATE_RENEWAL_INTERVAL_IN_SECONDS * 1000) || (1000 * 60 * 60), // 1 hour`

in https://github.com/jc21/nginx-proxy-manager/blob/4f10d129c20cc82494b95cc94b97f859dbd4b54d/backend/internal/certificate.js#L24

[...]

PR #3702, opened 2024-04-15
Set interval timeout for certificates via environment variable

MrDKOz:
There's no need to run certificate renewal every hour, making it at least configurable would at least allow people such as myself to configure this to a more suitable interval by defining the interval against the environment variable `CERT_INTERVAL_TIMEOUT` in milliseconds.

There's a further discussion about this over on #677

Note that with the current implementation using `setInterval` ([docs](https://developer.mozilla.org/en-US/docs/Web/API/setInterval)) there are limits to the maximum value of [just shy of 25 days](https://stackoverflow.com/questions/12633405/what-is-the-maximum-delay-for-setinterval) due to the value being a 32-bit signed integer.

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -29,7 +29,7 @@ function omissions() {
 const internalCertificate = {
 
 	allowedSslFiles:         ['certificate', 'certificate_key', 'intermediate_certificate'],
-	intervalTimeout:         1000 * 60 * 60, // 1 hour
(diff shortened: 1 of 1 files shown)


Jordibgzashtita:
is this bug fixed https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3580

MrDKOz:
> is this bug fixed #3580

No, this is something completely unrelated.

MrDKOz:
Closing this issue due to the timer not really having any use due to everything being reloaded once an hour - an oversight on my part. Apparently v3 is being worked on so maybe it'll be fixed then.

PR #3729, opened 2024-04-28
move advanced_config section of /app/templates/_location.conf to top of default config

clhey:
I found that custom configurations cannot override the default configurations because nginx prioritizes the configurations appearing earlier. Therefore, I moved the advanced_config section of /app/templates/_location.conf to top of default config。

config：
image

before nginx proxy file:
image

after nginx proxy file:
image

see this issue:
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/2675#issuecomment-2081407884

### DIFF
--- a/backend/templates/_location.conf
+++ b/backend/templates/_location.conf
@@ -1,4 +1,6 @@
   location {{ path }} {
+    {{ advanced_config }}
+
     proxy_set_header Host $host;
     proxy_set_header X-Forwarded-Scheme $scheme;
     proxy_set_header X-Forwarded-Proto  $scheme;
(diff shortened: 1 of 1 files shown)


whc2001:
Are you sure that the config entry on the top will take precedence in Nginx? I have tried myself and the result is opposite. 

image

image

The line where the red arrow points to has `proxy_set_header Host $host;` originally, and I was not able to get my override config of `$http_host` work until I manually delete the auto generated line below.

Erzbir:
Why do that? It is no work well if I want set Host header

dgm:
Unfortunately, @Erzbir , It seems to me that I get an error with two line lines setting the Host header, no matter which order they are.   I have to delete the $host one by hand in order to override the Host header in a custom location.   

Why would someone do this?

Server A  hosts a mail program, and server B embeds it  under b/mail/   ...
Or maybe you are in the process of migrating a service and need to intercept a partial path and send it somewhere else ...

Issue #46, opened 2019-01-08
Add GeoIP module


teodorch85:
Hi! As I am noob with nginx is it possible to enable the GeoIP module so we can limit access also by location?
https://docs.nginx.com/nginx/admin-guide/security-controls/controlling-access-by-geoip/

Thank you!

jc21:
Sounds doable. Makes sense to include with improvements in #38 as well.

wuast94:
the geoip module is active by default so no worry. its just a bit of path mapping and edit config files :P 

the main problem of geoip module is using legacy geoip databases that are deprecated.
from https://dev.maxmind.com/geoip/legacy/downloadable/
to https://dev.maxmind.com/geoip/geoip2/geolite2/ wich is using 
maxmind databases 

but the geoip module from nginx is using teh legacy databases. so i dont know if its an issue from nginx itselfs or from npm. it sounds nice to block countries etc but i thing nginx need to fix it first before this feature can addet by devs to the container or am i wrong ? 
i wrote more to -> #78

jc21:
Yep I've found the legacy databases are not longer supplied anymore so everyone has to convert to geo2.

Seems it's trivial to add the new module: https://docs.nginx.com/nginx/admin-guide/dynamic-modules/geoip2/

Also the geolite databases need to be baked in to the docker image.

wuast94:
I think it should be better to make this possible over mounting the nginx dir to host system. So everyone can edit this as wanted (use other modules or databases) 

And I think the geo2 module should be addet by nginx docker Mainter. Many modules are installed by defoult: https://github.com/nginxinc/docker-nginx/blob/master/stable/alpine/Dockerfile

jc21:
You can already do that if you want to. In any docker image there is no restriction in what you're mounting and where. Even though I bake in config or files you're always able to override them.

corvy:
I would really love this feature. :)

ghallford:
This would be awesome because in order to use [this](https://github.com/gilbN/geoip2influx)

Currently nginx fails with this error when I add the custom log format: 

> 2020/07/09 16:36:21 [emerg] 3022#3022: unknown "geoip2_data_city_name" variable

I need the geoIP information to send to InfluxDB custom format:

[...]

PR #3766, opened 2024-05-19
Add include for `root_top.conf` in the nginx.conf

kroegerama:
This pull request allows custom configuration of the root config in the top of the file. This can be used to load modules, which is not possible at the end of the config file. There is already a `http_top.conf`, so `root_top.conf` is a logical addition.

The PR NginxProxyManager/docker-nginx-full#28 added the geoip2 module as requested in Issue #46. But currently there is no way to actually load the modules if you cannot mount custom files to `/etc/nginx/modules/` like in the Truenas Scale NPM App.

Now you can just add a `root_top.conf` to the persistent mounted `/data/custom` folder and add the following content:
```
load_module /usr/lib/nginx/modules/ngx_http_geoip2_module.so;
load_module /usr/lib/nginx/modules/ngx_stream_geoip2_module.so;
```

### DIFF
--- a/docker/rootfs/etc/nginx/nginx.conf
+++ b/docker/rootfs/etc/nginx/nginx.conf
@@ -14,6 +14,9 @@ error_log /data/logs/fallback_error.log warn;
 # Includes files with directives to load dynamic modules.
 include /etc/nginx/modules/*.conf;
 
+# Custom
+include /data/nginx/custom/root_top[.]conf;
+
 events {
 	include /data/nginx/custom/events[.]conf;
 }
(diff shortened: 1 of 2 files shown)


jc21:
These custom includes are getting wild.

Can you please update the documentation with this addition as well:
https://nginxproxymanager.com/advanced-config/#custom-nginx-configurations

This file can be found at `docs/src/advanced-config/index.md`

kroegerama:
Thank you so much for your fast response! I updated the documentation and added a small snippet on how to enable the geoip2 module.
If you don't want the snippet in the documentation, let me know so I can remove it.

Issue #1862, opened 2022-02-15
Let's Encrypt: DNS Challenge for Cloudflare not working (conflicting dependencies)


MBennGit:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
Tried to add a certificate via DNS Challenge for Cloudflare domain. After starting the process, a timeout error appears. If the propagation seconds are set, an error is given hinting at conflicting dependencies with the `certbot-dns-cloudflare` pip package.

**Nginx Proxy Manager Version**
v2.9.15

**To Reproduce**
1. Go to the tab "SSL Certificates"
2. Click on "Add SSL Certificate"
3. Enter the domains "*.example.com, example.com"
4. Select "Use DNS Challenge", Cloudflare, and set API Key
4. Set Propagation Seconds (450 Seconds) (Optional)

**Expected behavior**
A SSL Wildcard Certificate is created.

**Screenshots**

**Operating System**
Raspberry Pi - Raspbian GNU/Linux 11 (bullseye)
docker-compose version 1.29.2

**Additional context**

[cloudflare-dns.log](https://github.com/NginxProxyManager/nginx-proxy-manager/files/8075176/cloudflare-dns.log)

tkjouu:
I have the same problem.

Jiriteach:
Just discovered this after much investigation. Related to #1769

MBennGit:
> have you created an api key with dns edit for all zones or a single zone?
> 
> have you tried making a new key?

Yes, created a new key, edit for all zones. Tested the key with the curl command given by cloudflare.

MBennGit:
> Just discovered this after much investigation. Related to #1769

I will test the recommended fixes and report back.

cryd-s:
is the latest version now creating the certificates? Cause mine send me an internal error :(

wildernessfamily:
I was having the same problem. I tired about six months ago and received the same error. I had some time this weekend to try to get it up and going. Again, I received the same error. I realized what it was. When you paste in your API key, if you look very careful, prior to the equal sign "=" there is a space. Remove the space and you should be good to go. I zoomed in on @cryd-s image, and it appears to have the same space. Hope that helps :)

ehtishammubarik:
[...]

PR #3781, opened 2024-05-29
Remove spaces around Cloudflare API Credential

jinhei:
I've run into [this same issue](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1862#issuecomment-1599770115) multiple times, as the common practice is to simply replace the placeholder token. However, doing so is incorrect syntax.

Proposal is to remove spaces around the `=` such that overwriting the placeholder API key with the user's "just works".


BluCobalt:
This same issue has given me headaches before. This should definitely be added.

PR #3792, opened 2024-06-05
Update 30-ownership.sh to make chown optional

ian351c:
Make setting file permissions optional with SKIP_FILE_OWNERSHIP environment variable.

### DIFF
--- a/docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/50-ipv6.sh
+++ b/docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/50-ipv6.sh
@@ -10,6 +10,7 @@ log_info 'IPv6 ...'
 
 # Lowercase
 DISABLE_IPV6=$(echo "${DISABLE_IPV6:-}" | tr '[:upper:]' '[:lower:]')
+SKIP_FILE_OWNERSHIP=$(echo "${SKIP_FILE_OWNERSHIP:-}" | tr '[:upper:]' '[:lower:]')
 
 process_folder () {
 	FILES=$(find "$1" -type f -name "*.conf")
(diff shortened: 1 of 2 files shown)


ian351c:
This is so that I can use NPM in a VM that has "shared folders".

docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/50-ipv6.sh
```diff
@@ -31,9 +32,17 @@ process_folder () {
 		echo "$(sed -E "$SED_REGEX" "$FILE")" > $FILE
 	done
 
-	# ensure the files are still owned by the npm user
-	chown -R "$PUID:$PGID" "$1"
+
+		# ensure the files are still owned by the npm user
+		chown -R "$PUID:$PGID" "$1"
```
jc21, line 37: Only this line in this file should be wrapped in the IF statement. You still want the SED above to function.

rrolla:
Also there is a big problem with mariadb

On logs it shows that cannot create table, but that because of file ownership, mariadb cannot access those files, that live in `/var/lib/mysql`
```bash
app_1  | [6/8/2024] [2:40:17 AM] [Global   ] › ✖  error     create table `migrations` (`id` int unsigned not null auto_increment primary key, `name` varchar(255), `batch` int, `migration_time` timestamp) - ER_CANT_CREATE_TABLE: Can't create table `npm`.`migrations` (errno: 13 "Permission denied")

```

example docker-compose.yml

```yml
  app:
    image: 'jc21/nginx-proxy-manager:2.11.2' 
    ...
    environment:
      PUID: 1000
      PGID: 1000
      DB_MYSQL_HOST: "db"
      DB_MYSQL_PORT: 3306
      DB_MYSQL_USER: "npm"
      DB_MYSQL_PASSWORD: "npm"
      DB_MYSQL_NAME: "npm"
    volumes:
      - ./data:/data

  db:
    image: 'jc21/mariadb-aria:10.11.5'
    environment:
      MYSQL_ROOT_PASSWORD: 'npm'
      MYSQL_DATABASE: 'npm'
      MYSQL_USER: 'npm'
      MYSQL_PASSWORD: 'npm'
      MARIADB_AUTO_UPGRADE: '1'
    volumes:
      - ./data/mysql:/var/lib/mysql
```

On start application Setting ownership by envs PUID: 1000 PGID: 1000 also for ./data/mysql directory ant thats why mysql is not working anymore, because it runs as mysql user

As dirty workaround, to at least nginx proxy manager starts and works is to 
```bash
# after running docker compose up need to change owner for mysql
docker exec -it nginx-proxy-manager_db_1 chown -vR mysql /var/lib/mysql/
```

jc21:
Yeah the `mysql` folder should not be within the `data` folder for this very reason.

The [documentation](https://nginxproxymanager.com/setup/#using-mysql-mariadb-database) doesn't do this either (anymore)

crazychatting:
skipping the ownership would held to include custom configs via docker swarm config mounts!

crazychatting:
👍

PR #3819, opened 2024-06-22
force-ssl behind another proxy using http_x_forwarded_proto

paucapo:
Fix for issue #3365 and pull request #3589

### DIFF
--- a/docker/rootfs/etc/nginx/conf.d/include/force-ssl.conf
+++ b/docker/rootfs/etc/nginx/conf.d/include/force-ssl.conf
@@ -2,8 +2,8 @@ set $test "";
 if ($scheme = "http") {
 	set $test "H";
 }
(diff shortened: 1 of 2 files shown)


jc21:
Change looks good, can you confirm that this docker image works as you expect?

paucapo:
Seems like is working as expected, but maybe some additional tests with configurations a bit more elaborated than mine will be a good idea.

cbenard:
This doesn't work when you have Cloudflare Tunnel (cloudflared) connecting to an https nginx endpoint and they requested HTTP.

The working code changes this:

```
if ($http_x_forwarded_proto = "https") {
	set $test "";
}
```

to this:

```
if ($http_x_forwarded_proto = "http") {
	set $test "H";
}
```

It then works as expected.

PR #3876, opened 2024-07-15
Feature: PiHole Plugin

luca910:
That far on creation of a new Proxy host a new DNS record is created in PiHole.
TODO: Update Method
TODO: Delete Method

The plugin requires new env vars for the npm container:
PIHOLE_IP
PIHOLE_PASSWORD

### DIFF
--- a/backend/internal/PiHoleDNSPlugin.js
+++ b/backend/internal/PiHoleDNSPlugin.js
@@ -2,13 +2,17 @@ const axios   = require('axios');
 const cheerio = require('cheerio');
 const qs      = require('querystring');
 
-
-const PIHOLE_PASSWORD      = process.env.PIHOLE_PASSWORD;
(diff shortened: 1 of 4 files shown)


luca910:
Update method:
- delete
- create

PiHole has not implemented a method to update local dbs records

luca910:
bruh

luca910:
The plugin can be enabled and configured with docker compose env vars.


Help would be appreciated since this plugin is enabled globally and not per proxy host entry.
Maybe via an extra radio button. 

On the other hand, this may not be required since, outside the network, another DNS takes over with the public IP entry, if the domain is already existing and configured.

CrazyWolf13:
That actually sounds awesome!

Hoping for a merge :)

szolen:
Any information on when this feature will be released?

m4rkireland:
No plan to release this?

PR #3887, opened 2024-07-19
Update api.swagger.json

Sander0542:
The api.swagger.json file is outdated and it not in a valid OpenAPI format. I have created a simple tool which parses the models and generates an OpenAPI spec which is valid.

The tool I created and used to generate the file: https://github.com/Sander0542/nginx-proxy-manager-openapi


jc21:
This PR is no longer valid with recent changes.

PR #3907, opened 2024-07-31
Add rockenstein AG DNS Plugin

rag-franky:
Hi,

here is an addition with the DNS certbot plugin for the DNS services of the german full-service-provider rockenstein AG.

It would be nice if this was included in the next version. I tested the plugin by building the nginxproxymanager locally and copying the .js files into the official Docker container.

Let me know if there's anything else you'd like me to change.

Thank you!


rag-franky:
Perfect! It worked on the first try. 

I really ❤️ NginxProxyManager. 

That smells like ☕️! :-) 

Thank you very much!

Issue #3267, opened 2023-10-17
Real IP addresses in the log?


mxm199:
Good day to everyone, looking for help, tips

In short, there is such a scheme of the home network
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/133875469/fdfcb883-bd1d-4f5b-b6bb-a82fdafadcd6)
There is nothing between the router and NPM, just a direct port forwarding to the container with the NPM (192.168.0.200 80, 443).

Used docker-compose -
```
version: '3.8'
services:
  app:
    image: 'jc21/nginx-proxy-manager:latest'
    restart: unless-stopped
    container_name: nginx-proxy-manager
    ports:
      - '80:80'
      - '443:443'
      - '4488:81'
    environment:
      DB_HOST: "db"
      DB_PORT: 5432
      DB_USER: $(DB_USER)
      DB_PASSWORD: $(DB_PASSWORD)
      DB_NAME: $(DB_NAME)
      DISABLE_IPV6: 'true'
      X_FRAME_OPTIONS: "sameorigin"
    volumes:
      - ./data:/data
      - ./letsencrypt:/etc/letsencrypt
    depends_on:
      - db
    healthcheck:
      test: ["CMD", "/bin/check-health"]
      interval: 10s
      timeout: 3s
    networks:
      dockernet:
        ipv4_address: 172.16.0.2
  db:
    image: 'postgres:13'
    restart: unless-stopped
    container_name: nginx-proxy-manager-db
    environment:
      POSTGRES_USER: $(POSTGRES_USER)
      POSTGRES_PASSWORD: $(POSTGRES_PASSWORD)
      POSTGRES_DB: $(DB_NAME)
    volumes:
      - pg-data:/var/lib/postgresql/data
    networks:
      dockernet:
        ipv4_address: 172.16.0.3
volumes:
  pg-data:
networks:
  dockernet:
    external: true
```

A network has been created
`docker network create --subnet=172.16.0.0/24 dockernet`

Hosts have been added to NPM, everything works fine, full Internet access to all resources

To view logs, I use goaccess (xavierh/goaccess-for-nginxproxymanager). But the whole problem is that only one client address appears everywhere in the log - 172.16.0.1 (takes information (parses) for display directly from the npm logs folder).
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/133875469/e3c39444-3fe1-46d8-bf1c-46d8340878ed)

[...]

Issue #1358, opened 2021-08-28
X-forwared-for shows Cloudflare IP instead of real IP


bernikr:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
When using Cloudflare infront of NPM, NPM populates the `X-forwarded-for` header with the cloudflare IP instead of the real IP that cloudflare provices in its `X-forwarded-for` header. I would expect NPM to check if the source IP is one of cloudflare's and trust the `X-forwarded-for` header in that case. This seemd to be a feature in `2.6.0` but broken in all verisons afterwards.

**Nginx Proxy Manager Version**
`2.9.8`

**To Reproduce**
Set up a cloudflare domain and let it point to NPM. If you now navigate to the site the server behind NPM will get one of cloudflare's IPs as the `X-forwarded-for` header instead of the client IP

**Expected behavior**
Show the real client IP that cloudflare reports in the `X-forwarded-for` header.

**Operating System**
Synology NAS

**Additional context**
The closed issues #112 and #811 both suggest using the following settings as a workaround:

```
real_ip_header CF-Connecting-IP;
#real_ip_header X-Forwarded-For;
```

This is however a **major security risk** as it doesnt check if the request is actually coming from cloudflare. It allows anyone with knowledge of your IP to send requests pretending to be from cloudflare with an `CF-Connecting-IP` header set to any IP they want, (including lan ips such as 10.0.0.1) possibly circumventing any IP based security measures.

schevenin:
I opened a similar issue about this awhile ago. If you want to get this to work you'll have to use v2.6.0. Check this out: #811

schevenin:
Use v2.6.0 is what I found in my #811

TritonB7:
Found the following:

nginx configuration file snippet:
```
set_real_ip_from    03.21.244.0/22;
set_real_ip_from    103.22.200.0/22;
set_real_ip_from    103.31.4.0/22;
set_real_ip_from    104.16.0.0/13;
set_real_ip_from    104.24.0.0/14;
set_real_ip_from    108.162.192.0/18;
set_real_ip_from    131.0.72.0/22;
set_real_ip_from    141.101.64.0/18;
set_real_ip_from    162.158.0.0/15;
[...]

Issue #1230, opened 2021-07-11
nginx stream real ip


penalte:
Hello when i use the Stream feature  it works well but for game servers the ip of the client is the ip of the machine.

I found this not sure if it is relevant.
http://nginx.org/en/docs/stream/ngx_stream_realip_module.html

Thank you for this amazing software.

baudneo:
this is in the nginx.conf for NPM ->
`

    # Real IP Determination
	# Local subnets:
	set_real_ip_from 10.0.0.0/8;
	set_real_ip_from 172.16.0.0/12; # Includes Docker subnet
	set_real_ip_from 192.168.0.0/16;
	# NPM generated CDN ip ranges:
	include conf.d/include/ip_ranges.conf;
	# always put the following 2 lines after ip subnets:
	real_ip_header X-Real-IP;
	real_ip_recursive on;

`
and the ip_ranges.conf gets updated with cloudflare ips, not sure if there is a way to add some yourself and it be persistent?

penalte:
Can someone help with this, i really need the client ip in stream.

PR #3929, opened 2024-08-13
Switch from X-Real-IP to X-Forwarded-For in nginx.conf

### LINKED ISSUE #3582: Cloudflare ipv4 not loaded

woliver99:
X-Forwarded-For is more commonly used across various proxies and load balancers, including Cloudflare, which does not use X-Real-IP.

This should fix #3582 #3267 #1358 #1230

### DIFF
--- a/docker/rootfs/etc/nginx/nginx.conf
+++ b/docker/rootfs/etc/nginx/nginx.conf
@@ -66,7 +66,7 @@ http {
 	# NPM generated CDN ip ranges:
 	include conf.d/include/ip_ranges.conf;
 	# always put the following 2 lines after ip subnets:
-	real_ip_header X-Real-IP;
+	real_ip_header X-Forwarded-For;
 	real_ip_recursive on;
 
 	# Custom


woliver99:
Just realized that we could make it a variable so that way people could change it to whatever they want. But I still believe we should make the default x-forwarded_for instead of x-real-ip

cbenard:
I'm having to mount a custom `nginx.conf` to fix this. I would like this changed too.

jmarmstrong1207:
I believe that X-Real-IP may be used instead of X-Forwarded-For in the case that a non-trusted proxy spoofs the IP in X-Forwarded-For. X-Real-IP would always be the most recent IP

This is from my research when I had this issue with cloudflare proxies making NPM output incorrect IP in the logs

jkbudde:
`X-Forwarded-For` will list all IPs for any proxy the request has been through and including the client IP address provided to the original proxy (Cloudflare or otherwise). This header is capable of being spoofed which is the importance of having the trusted IPs available to nginx through `set_real_ip_from`. _There also is a proposed standard regarding the Forwarded HTTP Extension, see [RFC 7239](https://datatracker.ietf.org/doc/html/rfc7239)._

With Cloudflare, the source of truth is from the [`CF-Connecting-IP`](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ip) for domains with proxying enabled as noted by @jmarmstrong1207. For IPv6, the true IPv6 address will be in [`CF-Connecting-IPv6`](https://developers.cloudflare.com/fundamentals/reference/http-request-headers/#cf-connecting-ipv6). I use `CF-Connecting-IP` unless I know the service/host is not proxied through Cloudflare.

An option in the user interface to list all trusted IPs (beyond those already sourced) and to set the `real_ip_header`. Creating a variable where it defaults to either `X-Forwarded-For` or `X-Real-IP` may be beneficial if the approach may include the option for customization from end users of NPM in the future. _Of note, other service providers also have a similar header to Cloudflare, where having an option for directly customizing this would be beneficial in addition to sourcing Cloudflare and AWS known IPs._

medram:
Hello everyone, I run into the same issue, and I solved it like this:
```yml
npm:
    image: jc21/nginx-proxy-manager:2.12.1
    ports:
      - "80:80"
      - "443:443"
      - "8090:81"
    environment:
     ...
    volumes:
      - ${NFS_DOCKER_VOLUMES}/npm_volume/data:/data
      - ${NFS_DOCKER_VOLUMES}/npm_volume/letsencrypt:/etc/letsencrypt
      # Fixed by mounting it to this path.       
      - ./configs/server_proxy.conf:/data/nginx/custom/server_proxy.conf
```

My config file `./configs/server_proxy.conf`:

```txt
# Required to pass the real client IP address to the backend services 
real_ip_header CF-Connecting-IP;
```

Note: You can change `CF-Connecting-IP` with anything you want if needed.

Issue #2516, opened 2023-01-03
Support other ports (non 80/443) configuration in headers via UI settings


GAS85:
**Is your feature request related to a problem? Please describe.**
If my Application runs on different ports, e.g. `907X` it will cause a bunch of trouble - port will not be forwarded to the service and at the end redirect happens on port `80` or `443` that are not configured by npm.
Docker-compose
```yaml
version: "3.6"
services:
  nginx-proxy-manager:
    image: jc21/nginx-proxy-manager:latest
    container_name: npm
    environment:
      - PUID=${PUID}
      - PGID=${PGID}
      - TZ=${TZ}
      - DB_SQLITE_FILE=/config/database.sqlite
      - DISABLE_IPV6=true
    ports:
      # Public HTTP Port. Port Forwarding on Router is ON.
      - 9070:80
      # Public HTTPS Port. Port Forwarding on Router is ON.
      - 9071:443
      # Admin Web Port. Port Forwarding on Router is OFF. Internal Home Network Access only - 192.168.89.254:81.
      - 9072:81
    volumes:
      - ${WORKINGDIR}/docker/npm/config:/config
      - ${WORKINGDIR}/docker/npm/letsencrypt:/etc/letsencrypt
      - ${WORKINGDIR}/docker/npm/data:/data
```
![изображение](https://user-images.githubusercontent.com/6813635/210354844-69443697-f9e7-459b-b688-ceb9cb91d89f.png)

On a 1st line Service was found and it tells us to follow url `http://jackett.fritz.box:9070/UI/Dashboard`.
On a 2nd line Service would like to authorize user, but do not know port, so it skip it in a answer and ended up with wrong URL (no port) `http://jackett.fritz.box/UI/Login?ReturnUrl=/UI/Dashboard` and 404.

**Describe the solution you'd like**
Configure port in a `proxy_set_header Host $host:PORT;` will solve this issue.

![изображение](https://user-images.githubusercontent.com/6813635/210356619-0e2202f3-45d5-42fb-980b-6fe5127a99e8.png)

Unlike current solution it is only `proxy_set_header Host $host;`. Add Possibility to enable and add custom port to the header Host.

**Describe alternatives you've considered**
I tried to use "Custom Location" and set it to the root so I can add `proxy_set_header Host $host:PORT;` but this will cause this entry twice: without port and at the end with port that cause an error.

Alternative 2 is to mount `proxy.conf` to the host machine:
```yaml
    volumes:
[...]

PR #3945, opened 2024-08-21
Modify the Host value to $http_host in order to pass the port number …

1oca1h0st:
…while maintaining transparency.

Modify the Host value from $host to $http_host in order to pass the port number while maintaining transparency.

#2516 related.


1oca1h0st:
it seems is a network error, please try merge again.

PR #3973, opened 2024-09-04
Add set directives for proxied paths to keep nginx from crashing if upstream is down

ddshd:
If the upstream is down on a `/path` setting on a proxy setting it will cause nginx to crash. Adding `set` directives pointing to the upstream information allows nginx to startup even if the upstream is down and throw HTTP errors instead.


horsley:
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4103 @ddshd

Issue #3287, opened 2023-10-26
ER_NOT_SUPPORTED_AUTH_MODE: Client does not support authentication protocol requested by server; consider upgrading MySQL client


z499738316:
Hello, I encountered an error connecting to MySQL while installing nginx proxy manage
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/ddbd2685-209e-40aa-9937-da24a2c08364)
I installed MySQL 8.0 in Docker，But nginx proxy management told me that the connection was wrong. Let me know
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/f372fb56-7225-4c01-add1-696f38781598)
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/08f71f42-9e85-4fe0-bf0c-8769a0dd59b8)
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/a864c6b6-13ac-4e01-8aa2-6eada7001e81)
Can you help me check the errors? MySQL and nginx are both in the same Docker service
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/6f5e7123-8882-4baa-a3c6-ae01847dbd40)

z499738316:
The network is functioning normally
![image](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/37020005/d0065548-f842-4769-9cc0-5848b00ac020)

DylanManiatakes:
Im having this same issue, did you ever get it resolved

z499738316:
> Im having this same issue, did you ever get it resolved

不要使用mysql8.0以上版本，需要使用mysql5.7版本

ggbr:
It seems like you have to update the Mysql client lib

ggbr:
I followed this and it resolved for Mysql 8.0.36

```
Execute the following query in MYSQL Workbench

ALTER USER 'root'@'localhost' IDENTIFIED WITH mysql_native_password BY 'password';

Where root as your user localhost as your URL and password as your password

Then run this query to refresh privileges:

flush privileges;

Try connecting using node after you do so.

If that doesn't work, try it without @'localhost' part.
```

Link:
https://stackoverflow.com/questions/50093144/mysql-8-0-client-does-not-support-authentication-protocol-requested-by-server

gitown:
> I followed this and it resolved for Mysql 8.0.36
[...]

PR #3991, opened 2024-09-10
swap mysql library and knex client for mysql2

nlynzaad:
Since MySQL 9 the use of the native_password plugin is no longer supported and caching_sha2_password is now required.

Currently npm makes use of the [mysql](https://www.npmjs.com/package/mysql) library. This library unfortunately does not support the caching_sha2_password authentication method.

[mysql2](https://www.npmjs.com/package/mysql2) supports the new caching_sha2_password methods and "is mostly API compatible with [Node MySQL](https://github.com/mysqljs/mysql) and supports majority of features"

This PR changes the mysql library for mysql 2 and updates the knex configs to use mysql2. This closes issue #3287 

Basic tests against a new and existing npm database (updated to mysql 9) worked flawlessly. 

I might be doing something wrong, but I was not able to get the scripts/test_dev.sh to run as it was seeking a taskfile in the dev container.


jc21:
Thanks for testing :)

Issue #3965, opened 2024-08-29
WTF :login 502


peterwillcn:
WTF : login 502

```
  ngx:
    image: 'docker.io/jc21/nginx-proxy-manager:latest'
    container_name: nginx
    restart: unless-stopped
    ports:
      - '80:80'
      - '81:81'
      - '443:443'
    volumes:
      - /data/nginx:/data
      - /data/nginx/letsencrypt:/etc/letsencrypt
   ```

![image](https://github.com/user-attachments/assets/1865f6e2-0265-4e3d-a9e8-4aa1a1043c8e)

ip520011:
我也是，目前还没解决办法

riruigit:
> 找到解决办法了，我发现是因为我的宿主机上装了node 导致的，把 node 卸载掉，就可以了。 重要的事情说三遍！ 宿主机必须删除 node 环境！宿主机必须删除 node 环境！宿主机必须删除 node 环境！

他这个是基于docker运行的，应该跟你这个没多大关系，我也是docker 复现了。

riruigit:
找到解决方案了吗？

ip520011:
> > 找到解决办法了，我发现是因为我的宿主机上装了node 导致的，把 node 卸载掉，就可以了。 重要的事情说三遍！ 宿主机必须删除 node 环境！宿主机必须删除 node 环境！宿主机必须删除 node 环境！
> 
> 他这个是基于docker运行的，应该跟你这个没多大关系，我也是docker 复现了。

那就不清楚了，我复现了好几次，删除了 node 环境，并且重启服务器，就可以了

laopaoer-wallet:
跟node没关系。。。。

icerao:
I have same issue.start log I found:
`[9/12/2024] [4:58:36 AM] [Setup    ] › ℹ  info      Creating a new user: admin@example.com with password: changeme
[9/12/2024] [4:58:37 AM] [Setup    ] › ℹ  info      Initial admin setup completed
[9/12/2024] [4:58:37 AM] [Setup    ] › ℹ  info      Default settings added
[9/12/2024] [4:58:37 AM] [Setup    ] › ℹ  info      Logrotate Timer initialized
[9/12/2024] [4:58:37 AM] [Global   ] › ⬤  debug     CMD: logrotate /etc/logrotate.d/nginx-proxy-manager
[...]

Issue #3585, opened 2024-02-26
Cannot access admin interface on v2.11.1 (502 Gateway Error)


krosseyed:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
I cannot log into the admin interface after deploying a fresh v2.11.1 container on a Mikrotik CHR on AWS using a t3.micro.

**To Reproduce**
1. [Setup container support](https://help.mikrotik.com/docs/display/ROS/Container) and [enable device-mode on AWS](https://forum.mikrotik.com/viewtopic.php?p=1058637#p1058637) for Mikrotik CHR
2. [Create config.json](https://old.reddit.com/r/nginx/comments/pt352s/bad_gateway_on_login/hdvucab/) and move it to /npm on CHR
3. Install container using the following config

```
/interface/veth
add name=veth1 address=172.17.0.2/24 gateway=172.17.0.1

/interface/bridge
add name=docker

/ip/address
add address=172.17.0.1/24 interface=docker

/interface/bridge/port 
add bridge=docker interface=veth1
/ip/firewall/nat/add chain=srcnat action=masquerade src-address=172.17.0.0/24 

/container envs
add key=DB_SQLITE_FILE name=npm_envs value=/data/database.sqlite
add key=DEBUG name=npm_envs value=true

/container mounts
add dst=/data name=/npm/data src=/npm/data
add dst=/etc/letsencrypt name=/npm/letsencrypt src=/npm/letsencrypt
add dst=/app/config/production.json name=/npm/config.json src=/npm/config.json

/container
add envlist=npm_envs interface=veth1 logging=yes mounts=/npm/data,/npm/letsencrypt,/npm/config.json root-dir=/npm/root start-on-boot=yes workdir=/app
```
4. Remove /etc/nginx/conf.d/dev.conf by going into the container and deleting it
```
/container shell 0
rm /etc/nginx/conf.d/dev.conf
```
5. Shutdown and Startup the container and wait about 10 minutes to start up. Then access via http://<IP_OF_AWS_INSTANCE>:81 using the default admin credentials

**Expected behavior**
To be able to log into the admin console. Not get a 502/Bad Gateway error

**Screenshots**
[...]

Issue #3574, opened 2024-02-21
Getting 502 Bad Gateway error


nidhin077:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - No
- Are you sure you're not using someone else's docker image?
  - Yes 
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
I'm getting a 502 Bad Gateway error when adding a location in the nginx proxy manager. Could you please help? 

**Nginx Proxy Manager Version**
v2.10.4

**To Reproduce**
Use a v2.10.4  version image and try to add the location

lilws:
I have just upgrade to v2.11.3 and got the same problems, but it is only happen to 1 specific host. Other is running fine. But rollback to 2.10.4 eveything back to normal.

My situation:

I have a network A: 116.78.110.16 - Home 1
I have a network B: 172.25.10.26 - Home 2 in another place

All the NPM setup was running on Network A and eveything is fine with this.
In the network B I have 2 hosts through NPM in Network A, and this is where the problems happen when upgrade to 2.11.x.
I can access the first host in network B without any problems, then the second host in network B always return 502 Bad Gateway.

There is no error or log so it seems I can't track what the problems here.

lilws:
I have just upgrade to v2.11.x and got the same problems 502 Bad gateway, but it is only happen to 1 specific host. Other hosts is running fine. But rollback to 2.10.4 eveything back to normal.

My situation:

I have a network A: LAN - 192.168.1.1 and internet - 116.x.x.x - Home 1
I have a network B: LAN - 192.168.1.1 and internet - 171.y.y.y - Home 2 in another place

All the NPM setup was running on Network A and eveything is fine with this.
In the network B I have 2 hosts through NPM in Network A - yes, I must forward these 2 ports in network B so A can forward to it, and this is where the problems happen when upgrade to 2.11.x.
I can access the first host in network B without any problems, then the second host in network B always return 502 Bad Gateway.

There is no error in log so it seems I can't track what the problems here because when test server reachability it works.

Issue #3157, opened 2023-08-24
Fetching IP Ranges Timeout


theflu:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes 
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
Long startup time due to waiting for "Fetching IP Ranges" to timeout.
```
nginx-proxy-manager | [8/24/2023] [7:04:29 PM] [IP Ranges] › ℹ  info      Fetching IP Ranges from online services...
nginx-proxy-manager | [8/24/2023] [7:04:29 PM] [IP Ranges] › ℹ  info      Fetching https://ip-ranges.amazonaws.com/ip-ranges.json
nginx-proxy-manager | [8/24/2023] [7:06:39 PM] [IP Ranges] › ✖  error     connect ETIMEDOUT 99.84.208.32:443
```

**Nginx Proxy Manager Version**
v2.10.4

**To Reproduce**
1. Require your docker server to use http proxy for internet access
2. Add proxy env  var to compose file IE "HTTP_PROXY=proxy.example.com:3128" "HTTPS_PROXY=proxy.example.com:3128"
3. Start container
4. See logs

**Expected behavior**
Use the http proxy in the environment variables or a way to specify the ip range in the compose file

**Operating System**
Ubuntu 22.04.3 LTS (GNU/Linux 5.15.0-79-generic x86_64)

NeDolM:
This issue even happens right out the box

NovaShen555:
and it still not be solved......

rer-coder:
Yes, it is still a problem in tencent cloud.

rer-coder:
I can get the https://ip-ranges.amazonaws.com/ip-ranges.json information with the `curl` command in the Docker container, but in NPM, I use the `fetch` command of node to execute the statement:
```
const fetch = require('node-fetch');
fetch('https://ip-ranges.amazonaws.com/ip-ranges.json').then(response => response.text()).then(data => console.log(data)).catch(error => console.error('Error fetching data:', error));
```
The return value is:
```
> Error fetching data: FetchError: Invalid response body while trying to fetch https://ip-ranges.amazonaws.com/ip-ranges.json: read ECONNRESET
    at PassThrough.<anonymous> (/app/node_modules/node-fetch/lib/index.js:400:12)
[...]

PR #4021, opened 2024-09-26
Fix the stuck at startup causing 502 bad gateway.

jqknono:
Some cloud vendors may have been blocked by AWS, this led to stuck at startup.

### DIFF
--- a/backend/index.js
+++ b/backend/index.js
@@ -15,7 +15,6 @@ async function appStart () {
 		.then(() => {
 			return apiValidator.loadSchemas;
 		})
-		.then(internalIpRanges.fetch)
 		.then(() => {
 
 			internalCertificate.initTimer();


jqknono:
There are some issues about 502 bad gateway, cannot login. 
The direct cause is the NPM getting stuck at startup, some of them related to the fetching of https://ip-ranges.amazonaws.com/ip-ranges.json

Some cloud vendors like Tencentcloud may block aws or be block by aws. Also , it may take a long time downloading this file, which blocked the port 81 backend.

https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3965
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3585
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3574
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3157

backend/index.js
```diff
@@ -15,7 +15,6 @@ async function appStart () {
 		.then(() => {
 			return apiValidator.loadSchemas;
 		})
-		.then(internalIpRanges.fetch)
```
jc21, line 18: Removing this is not good enough to fix the problem. This call might be in the wrong place but you haven't put a better one in the pr.

yangluoning:
@jc21 develop branch dont have your commit，Is it flushed by other commit?

jc21:
> @jc21 develop branch dont have your commit，Is it flushed by other commit?

I'm not sure what "commit" you're talking about? This PR is not mine and is not merged either.

jqknono:
For a quick workaround, you can run the following commands:

```bash
NPM_CTR_NAME=nginxproxymanager
docker exec $NPM_CTR_NAME sed -i 's/\.then(internalIpRanges\.fetch)//g' /app/index.js
docker restart $NPM_CTR_NAME
```
This will patch the application to skip fetching the IP ranges on startup.

Alternatively, as of [#4398](https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4398), you can use the IP_RANGES_FETCH_ENABLED environment variable to control this behavior.

To disable the fetch, set:

```env
IP_RANGES_FETCH_ENABLED=false
```
This will also prevent the application from hanging at startup. However, **note** that the default behavior is still to attempt downloading the IP ranges unless explicitly disabled.

backend/index.js
shiyoutaohua: This does solve part of the users’ problems. Could we consider allowing this operation to be skipped through an environment variable configuration?

PR #4047, opened 2024-10-10
Typo fixed in api.swagger.json

rbielawski:
I believe it's self-explanatory 🙃


jc21:
This file no longer exists after rework.

PR #4076, opened 2024-10-16
Fix schema validation errors

### LINKED ISSUE #4074: Strange errors in UI on 2.12..0
**Checklist**
  - Yes
  - Yes
  - Yes

**Describe the bug**
I'm receiving strange message "data/value must match exactly one schema in oneOf" across the UI on many pages. 

v2.12.0

**To Reproduce**
1. Go to 'Hosts > Redirection Hosts'
2. Click on 'Add redirection Host' and leave the scheme to 'Auto'
3. Fill in the values for 'domain names' and 'forward domain'
4. Click Save
5. See error
image

Another area where this error Persists is 'settings > default site'
1. Edit 'default site'
2. Try selecting '404 Page' or 'No Response (444)' 
3. Click Save
image

**Expected behavior**
Options should behave as expected.

**Screenshots**
Screenshots attached below repro steps. 

**Operating System**

**Additional context**
N/A

Nephiel:
See https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4074#issuecomment-2417649081
Should fix #4074.

### DIFF
--- a/backend/schema/components/redirection-host-object.json
+++ b/backend/schema/components/redirection-host-object.json
@@ -28,7 +28,7 @@
 		},
 		"forward_scheme": {
 			"type": "string",
-			"enum": ["http", "https"]
+			"enum": ["auto", "http", "https"]
 		},
 		"forward_domain_name": {
 			"description": "Domain Name",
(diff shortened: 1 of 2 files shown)


jc21:
Ah yes good catch. If the json schema wasn't coercing types (legacy requirement) then oneOf would have been fine.

PR #4086, opened 2024-10-19
Postgresql support

jeffreysjauwmook:
Hi I did some more extensive testing and fixed any issues I encountered.

To circumvent the issues with json fields I decided to cast the fields as nvarchar since the syntax is similar in both Mysql and Postgresql. The functionality  work similar when doing this on both Postgresql and Mysql.

It seems be working. However I haven't been able to test the certificates section yet. But the rest is working fine. 

Let me know if what you  guys think or if you have any comments on the changes.


jc21:
FYI I'm aware this exists and conflicts with #4087 but just need properly review both and decide on the best one.

jc21:
I've taken parts of this PR and merged it. Thanks for the contribution :)

PR #4087, opened 2024-10-19
Add Postgresql support

Dim145:
I have adapted most of the SQL queries using the ‘objection’ library as much as possible. 
These were tested in a containerised app with the Postgres and MySQL databases.

I haven't tested SSL certificate creation, but the queries in this section are normally compatible.

### DIFF
--- a/backend/internal/access-list.js
+++ b/backend/internal/access-list.js
@@ -81,7 +81,7 @@ const internalAccessList = {
 
 				return internalAccessList.build(row)
 					.then(() => {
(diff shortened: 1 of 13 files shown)


mokeev1995 (approved):

jc21:
FYI I'm aware this exists and conflicts with #4086 but just need properly review both and decide on the best one.

jc21:
I've taken parts of this PR and merged it. Thanks for the contribution :)

Issue #4011, opened 2024-09-20
proxy_set_header Host $http_host setting does not take effect


chung1912:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**

**Nginx Proxy Manager Version**

2.11.3

**To Reproduce**
After proxy_set_header Host $http_host is set, it will be placed at the front of "location", resulting in the configuration not being effective. For details, please see the figure
My configuration on the page is as follows:
![image](https://github.com/user-attachments/assets/f86d8bad-db29-4141-9fe8-8a16fee632c8)
The configuration information seen in conf is as follows:
![image](https://github.com/user-attachments/assets/3c246d40-8cff-4e7f-b992-d08266657395)
As shown in the figure, since `proxy_set_header Host $host;` is generated by default and is after `proxy_set_header Host $http_host;`, `proxy_set_header Host $host;` will take effect.

**Expected behavior**

Custom configuration should be placed at the end, not at the beginning, to ensure that the custom configuration can take effect normally, as shown in the following figure
![image](https://github.com/user-attachments/assets/9956df96-1aa5-44cf-8c2c-9c9d632dc996)

**Operating System**

**Additional context**

eliasmainetti:
Same problem

swensorm:
[...]

Issue #3981, opened 2024-09-06
make the posibility to choose the way the host and x-forward-x are set in _location.conf


mcarbonneaux:
**Is your feature request related to a problem? Please describe.**
actualy after commit https://github.com/NginxProxyManager/nginx-proxy-manager/commit/280bac8b430a3465f1fbc4d7e8d94e148ad740e7 you cannot fix `host` and `x-forward-xxx` header in advanced config.

because the advanced config in [_location.conf](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/backend/templates/_location.conf) template are now before default config and in default config you set `host` header like that (and some x-forward headers):
``` 
proxy_set_header Host $host;
```

this make impossible to use Nginx Proxy Manager  with application that control not only the `fqdn` but also the `port` used of the `host` header sended by nginx (if are behind nat port forwarding for example).

like her : https://github.com/esphome/issues/issues/4327 

**Describe the solution you'd like**
To preserve the `host` AND the `port` you MUST use `$http_host` nginx variable in place of `$host`:
https://serverfault.com/questions/706438/what-is-the-difference-between-nginx-variables-host-http-host-and-server-na

Because `http_host` is [http_HEADER](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_http_) and is value of host header sended by the navigator.

Is different than [`host`](https://nginx.org/en/docs/http/ngx_http_core_module.html#var_host) that are processed `host` by nginx that use listen port in place of the host header port.

Can you replace `host` by `http_host` by default or add the posibility to switch in option the way the template generate the host header ?

**Describe alternatives you've considered**
Revert commit https://github.com/NginxProxyManager/nginx-proxy-manager/commit/280bac8b430a3465f1fbc4d7e8d94e148ad740e7 that bloc the possibility to overide the host and x-forward-xxx headers.

Or add option to abel to choose the placement of the advanced option before or after default config.

Issue #3980, opened 2024-09-06
Commit "advanced config move to top of default confg" break the posibility to overiding host header


mcarbonneaux:
**Describe the bug**
 

the commit https://github.com/NginxProxyManager/nginx-proxy-manager/commit/280bac8b430a3465f1fbc4d7e8d94e148ad740e7 break the posibility to overide host header...

now is imposible to overide Host header in custom section like that:
```
proxy_set_header Host $http_host;
```

like her : https://github.com/esphome/issues/issues/4327#issuecomment-2211683256 

**Expected behavior**
hability to overide the host header and x-forward-*

jjunior776:
👍

jjunior776:
it's not just about the possibility to override host header but the whole advanced section (probably) gets overridden by whatever is default.

information-security:
Honestly, I am surprised that such an important issue hasn't been resolved for such a long time!

cyberbix:
I helped myself with linking the proxy.conf from container into the data dir:

    volumes:
      - ./data:/data
      - ./letsencrypt:/etc/letsencrypt
      - ./data/nginx/custom/my_proxy.conf:/etc/nginx/conf.d/include/proxy.conf

the custom my_proxy.conf reads:

add_header       X-Served-By $host;
proxy_set_header Host               $proxy_host;
#does not work with some destinations i.e. AVM Fritzbox
#proxy_set_header Host              $host;
proxy_set_header X-Forwarded-Scheme $scheme;
proxy_set_header X-Forwarded-Proto  $scheme;
proxy_set_header X-Forwarded-For    $proxy_add_x_forwarded_for;
proxy_set_header X-Real-IP          $remote_addr;
proxy_pass       $forward_scheme://$server:$port$request_uri;

In special cases there could be some if-else used to match a specific host.

gamba47:
> Honestly, I am surprised that such an important issue hasn't been resolved for such a long time!

Hi @information-security . If you're surprised, you could help by donating to the project or helping with a commit with the fix.

I'm like you with this problem, but I never will complain about the developer behind the project. That's useless.

See you!

information-security:
Hi @gamba47 
Good for you! Now that you don't complain, you may do exactly what you asked me to do. That would be a much better effort toward fixing your problem rather than complaining about my comment!

[...]

Issue #3678, opened 2024-04-04
Location directive ( location /{ ) causing proxy host to land at openresty landing page


cirnoscodingclass:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
When working under the `Advanced` tab of `Edit Proxy Host` I place the following addition to the config dialogue box
```
location / {
   proxy_set_header Host $http_host;
   proxy_set_header X-Forwarded-Host $http_host;
   proxy_set_header X-Forwarded-Port $server_port;
   proxy_set_header X-Forwarded-Proto $scheme;
   proxy_set_header X-Forwarded-For $remote_addr;
}
```
When this is added my sites redirects me to this screen 

I have also tried to use `=/` in front of the location directive but none of my changes appear to register. Also tried to add this configuration in the `Custom locations` tab but this shuts down the proxy host.

Adding the lines directly to the proxy_host .conf file causes the proxy_host to go offline.

**Nginx Proxy Manager Version**
2.11.1

**Expected behavior**
For the headers to be applied to the location directive without redirecting traffic to the openresty page.

**Screenshots**
![Screenshot from 2024-04-03 21-03-36](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/156504723/aad4948c-e5e5-4445-8e1e-0e4330beb0a6)

**Operating System**
Debian 12 Bookworm

**Additional context**
I am trying run [danbooru](https://github.com/danbooru/danbooru), the SSL certificate is DNS challenged by cloudflare
```
Docker Engine - Community
 Version:           26.0.0
 API version:       1.45
 Go version:        go1.21.8
 Git commit:        2ae903e
 ```

unixweb:
Same issue here 
-  image: 'jc21/nginx-proxy-manager:2.10.4'. Works 
 -  image: 'jc21/nginx-proxy-manager:2.11.0' Don't works

![Bildschirmfoto 2024-04-04 um 19 21 37](https://github.com/NginxProxyManager/nginx-proxy-manager/assets/12023025/fa4d0a90-e438-46e7-8cd1-d6466c5c26ef)
[...]

PR #4098, opened 2024-10-21
Duplicate advanced_config to both top and bottom

whc2001:
Trying to mitigate #3729 which caused #586 #1003 #2712 #2804 #3277 #3678 #3980 #3981 #4011


whc2001:
Actually, this does not work. No matter the statement `proxy_set_header Host $http_host;` appears in the beginning or the end, if `proxy_set_header Host $host;` exists it would not lead to the intended behaviour. Some more code will be needed to correct this behaviour

PR #3618, opened 2024-03-10
Enable PROXY procotol for proxy hosts and streams

snordmann:
As I see not too much progress in https://github.com/NginxProxyManager/nginx-proxy-manager/pull/3537, I decided to copy most of the code and (try to) fix the tests.
In this PR I also included the code necessary to enable the PROXY protocol for Streams.

Thank you @jwklijnsma for the previous work

PR #4105, opened 2024-10-23
Add support for proxy_protocol in proxy_hosts and streams

### LINKED ISSUE #1114: Support for PROXY Protocol
Is it possible to accept PROXY Protocol traffic from within the UI? 
If I go in and manually edit the files it works, but requesting certs resets the files so it's not really a viable alternative.

https://docs.nginx.com/nginx/admin-guide/load-balancer/using-proxy-protocol/

snordmann:
Closes #1114
Related To #1882
Related To #3537
Related To #3618

### DIFF
--- a/backend/internal/nginx.js
+++ b/backend/internal/nginx.js
@@ -156,7 +156,8 @@ const internalNginx = {
 						{ssl_forced: host.ssl_forced}, {caching_enabled: host.caching_enabled}, {block_exploits: host.block_exploits},
 						{allow_websocket_upgrade: host.allow_websocket_upgrade}, {http2_support: host.http2_support},
 						{hsts_enabled: host.hsts_enabled}, {hsts_subdomains: host.hsts_subdomains}, {access_list: host.access_list},
-						{certificate: host.certificate}, host.locations[i]);
+						{certificate: host.certificate}, {proxy_protocol_enabled: host.proxy_protocol_enabled},
+						{loadbalancer_address: host.loadbalancer_address}, host.locations[i]);
 
 					if (locationCopy.forward_host.indexOf('/') > -1) {
 						const splitted = locationCopy.forward_host.split('/');
--- a/backend/migrations/20241022221324_proxy_protocol.js
+++ b/backend/migrations/20241022221324_proxy_protocol.js
@@ -1,0 +1,56 @@
+const migrate_name = 'proxy_protocol';
+const logger       = require('../logger').migrate;
+
+/**
+ * Migrate
+ *
+ * @see http://knexjs.org/#Schema
+ *
+ * @param   {Object}  knex
+ * @param   {Promise} Promise
+ * @returns {Promise}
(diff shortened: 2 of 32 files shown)


snordmann:
@jc21 What does the requires-verification label mean? Is it that somebody should test this version manually by setting their tag to `nginxproxymanager/nginx-proxy-manager-dev:pr-4105`?

If so I am currently running this image an tested a few of the functions (both streams and proxy hosts work with and without the proxy procotol. Editing existing proxy_host and streams works (something that does need automated tests, because I only noticed my incomplete implementation when testing manually, not in the cypress tests/CI - but that is for another time).

Anyway, you probably shouldn't rely on my word alone, so maybe somebody else could verify my changes :)

jc21:
You're correct, in the absence of automated tests I do ask the community to verify changes, and I don't expect anyone to write those tests.

Also I need to manually test things myself and it's hard to find enough time to go through all of the PR's but over Xmas I intend to.

Coolicky:
Hi @snordmann 

Gave it a try pulled the image `nginxproxymanager/nginx-proxy-manager-dev:pr-4105`

First thing is that i cannot update existing settings. I'm getting the following js popup

Screenshot from 2025-01-14 19-25-00

Also if created from scratch with proxy_protocol, its always offline
image

Otherwise saw no regression in terms of other functions.

sansbyte:
I've been using the nginxproxymanager/nginx-proxy-manager-dev:pr-4105 image successfully for several months without any issues. I have both hosts using with and without Proxy Protocol enabled. I'm hoping this PR can be merged soon so I can benefit from other changes being added to the main release. Let me know if there's any information that I can provide that will help with verification.

Xmoe:
Hi, I am very interested in this PR! I have setup a second NPM instance using the PR and it works! My setup is:

`Internet <-> Gateway (proxy using streams) <-> Home-Server using NPM`

Testing with a small whoami service, I can now see the external IP on my server, whereas before it showed me the IP of the gateway. One thing I haven't quite figured out yet:

While at home, I have a DNS redirect so that any request from a device in my LAN goes directly to the home-server, without taking the route through the gateway.

Without internal redirect: `PC <-> Internet <-> Gateway <-> Home-Server`
With internal redirect: `PC <-> Home-Server`

Unfortunately, the services configured in NPM to accept the proxy protocol now won't load at all, because I don't go through a proxy which uses the protocol.

Can NPM accept proxied and un-proxied traffic at the same time? Or do I have to configure a second host for every service such that it can be also reached internally?

One workaround idea I have: setup another internal gateway which also uses the proxy protocol. But that does seems a bit over the top.

beatles1:
I'd really like to use this feature. Is the only option at the moment to use the Docker image from October? How do we go about getting this merged or another image made based on this PR and the other changes to develop?

DeusOpForge:
come on guys... please! this feature is a must!

PR #4075, opened 2024-10-16
fix(backend): Replace deprecated http2 directive

lusu007:
Fixes #4060 

This PR replaces the `listen ... http2` directive with the new and not deprecated directive `http2 on`. This prevents warnings in the log output.

PR #4106, opened 2024-10-23
http2 directive migration

dreik:
reducing noise in logs

N.B. to reduce the noise for existing hosts you'll have to edit every single host with http2 enabled and simply press save.

### DIFF
--- a/backend/templates/_listen.conf
+++ b/backend/templates/_listen.conf
@@ -5,11 +5,16 @@
   #listen [::]:80;
 {% endif %}
 {% if certificate -%}
-  listen 443 ssl{% if http2_support == 1 or http2_support == true %} http2{% endif %};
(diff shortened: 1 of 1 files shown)


jc21:
See #4075 this is already attempted, though is still WIP which is why I haven't looked at it too deeply

dreik:
@jc21 sorry i've missed the PR you've mentioned as I've simply tried to fix the issues for myself. Meanwhile I've just checked all the comments left by @vladwing and proud to say that my PR already covered everything mentioned there.

vladwing:
@jc21 you decide which one to accept, but this looks ready to me

jc21:
Looks good to me

DaanSelen:
So this warning is fixed in the upcoming version?

PR #4107, opened 2024-10-24
Update _access.conf

chrismaffey:
the pass_auth and satisfy_any properties and now boolean true/false, they do not == 1 so the switching in this template breaks


jc21:
Good catch

chrismaffey:
Cool.  That is the first time I have contributed code to an open source project!  I hope it is useful.

PR #4109, opened 2024-10-24
Update Dockerfile to use ghcr.io/letsencrypt/pebble

mcpherrinm:
The dockerhub letsencrypt/pebble image is deprecated and we will remove it in the future. This switches to the ghcr.io hosted version.


jc21:
The github image doesn't have a `arm/7` build, so this fails in our multibuild setup.

PR #4125, opened 2024-10-31
Improve test coverage

snordmann:
In this PR my goal is to test more endpoints and maybe even the frontend. This should make verification of future changes easier.


snordmann:
One thing that I noticed relating to the tests: The GET response for `/api/proxy-hosts` schema does not verify. The `certificate` property is not set on the returned proxy-hosts.

How should this be fixed? I see two options:
* (a) Edit the schema to make certificate an optional property
* (b) Adjust the backend to always include the certificate property

snordmann:
I will most likely not continue my work on this.
I don't like the high rate of repetition in the current testing approach.

And even then we don't really test whether frontend and backend are compatible (we would probably need real E2E tests or at least contract tests - though contract tests seem a bit overkill for a project this small).

Issue #4122, opened 2024-10-30
Allow Changing the Default Ports; Currently Blocks Host Networking


mtwo:
I've seen several issues in this repository that request the ability to change the default listen ports for HTTP and HTTPS traffic, however most of these are solvable by configuring the correct bridge ports in Docker or Kubernetes.

However, there is a use case for NPM that requires changing the ports within the container: host networking. [Host networking offers significant performance improvements over bridge networking](https://docs.docker.com/engine/network/drivers/host/), as the Docker runtime doesn't have to handle any of the incoming or outgoing network traffic from the application. It is often recommended for network proxies like Nginx for this reason, and there are [some compelling benchmarks of the performance differences](https://stackoverflow.com/questions/49023800/performance-issues-running-nginx-in-a-docker-container/49274333#49274333).

Unfortunately, Docker host networking removes the ability for Docker to remap ports, meaning that the default ports of the application inside of the container are used instead. This causes issues in Docker environments were ports 80 and 443 are already used by the host operating system or other applications, which is the case for TrueNAS and several others: https://github.com/truenas/apps/issues/734

Being able to set the Nginx listen and web UI ports via an environment variable or some other mechanism would allow users in any Docker environment to gain the performance benefits of using the host network driver.

HafizAhmedMoon:
👍

marcbennasarp:
👍

PR #4127, opened 2024-10-31
Support for chaning the HTTP and HTTPS port that hosts will listen on

snordmann:
I have seen a few issues that mention that changing the port that nginx listens on is important (especially when using hostNetwork).

I am not totally convinced that NPM should allow these use-cases but I was bored enough to implement it anyway.

Related Issue: #4122

### DIFF
--- a/backend/internal/nginx.js
+++ b/backend/internal/nginx.js
@@ -5,6 +5,20 @@ const config = require('../lib/config');
 const utils  = require('../lib/utils');
 const error  = require('../lib/error');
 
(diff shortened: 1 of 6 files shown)


snordmann:
@csbde could you please give more information what excactly went wrong?

My first guess: You are not using hostNeteork and the default ports of 80 and 443 are exposed instead of the defined ports. I could probably improve the documentation in that regard.

I will investigate on the weekend, when I have some spare time.

Issue #4159, opened 2024-11-13
Avatar not found


villcabo:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**

When I delete a user who has performed operations such as add host, ssl certificates.

**Nginx Proxy Manager Version**

v2.12.1

**To Reproduce**
1. Create a user with the user administrator
2. Login with the created user
3. Create a certificate or host
4. With the administrator user, delete the created user.
5. Go to host or certificates and you will see the error “Cannot read properties of null (reading ‘avatar’)”

**Expected behavior**

**Screenshots**

![image](https://github.com/user-attachments/assets/f0b5c33f-828d-4276-ad7c-b3ac016ef594)

**Operating System**

Ubuntu 24.04

**Additional context**

shalafi99:
It seems this behavior was described under #2041

kerstenremco:
Added PR #4166 for this issue

Issue #2041, opened 2022-05-05
Hosts of a deleted user break the host list


v3ss0n:
Re-issued #1038  Because it is not fixed , and owner of that issue is not reopening .

This annoying bug breaks NPM rendering anyone unable to change , very bad for production environment.
Please consider looking into it.

damianog:
I can confirm this issue, after deleting an old admin user I get the error: Cannot read properties of null (reading 'avatar')

![image](https://user-images.githubusercontent.com/2786682/170830172-3b312a5a-d356-4b33-a61a-1f6558a77722.png)

@v3ss0n  have you got a solution to restore the Hosts list?

v3ss0n:
Only way is to go in and fix using SQL.

damianog:
OK I managed to enable the user by:
update user set is_deleted = 0 where is_deleted = 1;
in mysql db

So the issue depend on the avatar icon on the host list?
![image](https://user-images.githubusercontent.com/2786682/170835102-338fc221-4bd3-476b-9bfd-fe9e0f4ce96b.png)

The one with red spot is the undeleted user now only disabled.

@jc21 I whant to delete a secondary administrator user, but there are some proxy-host owned by this one.
How could I change the owner than?

danielclough:
I was able to reset the user by:
 1. entering database
 2. using db
 3. `UPDATE user SET is_deleted=1;`
 4. wait for restart (or manually restart)
 5. `UPDATE user SET is_deleted=0;`
 6. wait for restart (or manually restart)
 7. log in 
 
 Upon logging in the deleted user is back along with their hosts.
 
 There is probably a better way.

v3ss0n:
Thats shouldn't be the way , its a critical bug. We need a fix.

wbloos:
> I was able to reset the user by:
> 
>     1. entering database
> 
>     2. using db
> 
>     3. `UPDATE user SET is_deleted=1;`
> 
>     4. wait for restart (or manually restart)
> 
>     5. `UPDATE user SET is_deleted=0;`
> 
>     6. wait for restart (or manually restart)
> 
>     7. log in
> 
> 
> Upon logging in the deleted user is back along with their hosts.
> 
> There is probably a better way.

That is dangerous, because it will restore the admin user with the default username and password!! (admin@example.com/changeme)

etfz:
Just encountered this very unfortunate issue. Was able to recover by assigning the affected entries to a different owner.

[...]

PR #4166, opened 2024-11-15
Fix entries of a deleted user break the UI

kerstenremco:
Removed user (owner object) isn't passed to EJS. Included fallback icon and text
Fix for #4159 #2041


kerstenremco:
Hi @villcabo,

I can't reproduce this. I've tried to setup a user and a host, delete that user, but the hostslist is still rendering.
I even delete the whole userrecord in the database.

Can you provide steps to reproduce this?

SCR-20241125-osjd

kerstenremco:
accidentally closed this PR, see #4187

villcabo:
@kerstenremco Thanks it's working

frontend/js/app/nginx/access/list/item.ejs
peterlinkpr, line 1: Please, What is the exact location of this file?

PR #4179, opened 2024-11-22
Return generic auth error to prevent user enumeration attacks

### LINKED ISSUE #3873: Login: No relevant user found / Invalid password
Would it be a good practice to change the login screen error prompt to a generic one? Now it shows you if you have correct email or not when you try to login. 

Currently it gives these:
Wrong email: No relevant user found
Wrong password: Invalid password

maybe just a generic "error" or "try again" prompt for both?

tametsi:
When authentication fails, the generic error message "Invalid email or password" is returned. Thereby, no information about the existence of the user is given.

This prevents user enumeration attacks because the response for both scenarios is identical.

Fixes #3873

### DIFF
--- a/backend/internal/token.js
+++ b/backend/internal/token.js
@@ -5,6 +5,8 @@ const authModel  = require('../models/auth');
 const helpers    = require('../lib/helpers');
 const TokenModel = require('../models/token');
 
+const ERROR_MESSAGE_INVALID_AUTH = 'Invalid email or password';
+
 module.exports = {
 
 	/**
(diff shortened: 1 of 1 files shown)


niri81:
Since this fixes [CWE-204: Observable Response Discrepancy](https://cwe.mitre.org/data/definitions/204.html), I look forward to this being merged relatively quick as it is only a minor change in the code base directly influencing the security of NPM

HALOTSCH (approved):
LGTM

Issue #914, opened 2021-02-23
413 Request Entity Too Large


nfacha:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image? **Yes**
- Are you sure you're not using someone else's docker image? **Yes**

- If having problems with Lets Encrypt, have you made absolutely sure your site is accessible from outside of your network? **Yes, but not related**

**Describe the bug**
File uploads trigger `413 Request Entity Too Large`
Setting the advanced config `client_max_body_size 100m;` takes no effect and the problem persists

**To Reproduce**
1. Upload a larger file on a website going thru the proxy, 413
2. Set an appropriate `client_max_body_size`
3. try to upload again
4. See a 413

**Expected behavior**
`client_max_body_size ` would take effect and allow the upload as it does with regular nginx

**Screenshots**

Ey3scr34m:
I'm having the same issue.

centralhardware:
same issue when try to upload iso to proxmox

ichbinder:
Hi, I have found a solution. you can enter client_max_body_size in the costome file server_proxy.conf, then it works. You can read [here](https://nginxproxymanager.com/advanced-config/#custom-nginx-configurations) where and how to create the file.

Mano-Liaoyan:
At the Edit Proxy Host window, click Custom locations and click the gear button, set the following attributes to an appropriate value
<img width="374" alt="image" src="https://user-images.githubusercontent.com/43143420/149541530-8da43c8c-68a0-4fd3-8721-798ccb4362ab.png">

jenny787:
> Hi, I have found a solution. you can enter client_max_body_size in the costome file server_proxy.conf, then it works. You can read [here](https://nginxproxymanager.com/advanced-config/#custom-nginx-configurations) where and how to create the file.

I went into NPM docker files, I added the server_proxy.conf file and added this line:
"client_max_body_size 0;"

It still doesn't work.  I also tried adding it to seafhttp like the picture above and that didn't work either.  I still get 413 error.

jeffryjdelarosa:
I have the same issue

jenny787:
I have now tried this in all sorts of ways, and it never works.  Nothing seems to change the NPM behavior.  I thought it was something to configure on Seafile's side, but I tried it with different webdav servers and all of them will return 413.  There is no way to fix this.  Seems like there is something hardcoded in NPM.

[...]

PR #4217, opened 2024-12-09
client_max_body_size setting via UI

dreik:
Attempting to close ancient issue #914 

Probably not the ideal solution, potential improvements are to follow nginx [docs](https://nginx.org/en/docs/http/ngx_http_core_module.html#client_max_body_size) and allow this settings on both 

1. server level
2. host level
3. location level

As of now it is _only_ on host level


dreik:
not stale

PR #4235, opened 2024-12-15
Update DomainOffensive certbot plugin

FabianK3:
### Summary

This PR replaces the current DomainOffensive (do.de) cert bot with the new [official version](https://github.com/domainoffensive/certbot-dns-domainoffensive).
The current [do certbot repo](https://github.com/georgeto/certbot-dns-do) has been forked by DomainOffensive and seems to be the official bot now.

### Changes

Basically i have replaced the current certbot-dns-plugins config with the new repository and python [package](https://pypi.org/project/certbot-dns-domainoffensive/).
Looking into the new repository of DomainOffensive, i don't see any new additions or changes, yet i think using the official certbot seems to be a good thing for future versions.

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -153,11 +153,11 @@
 	},
 	"domainoffensive": {
 		"name": "DomainOffensive (do.de)",
-		"package_name": "certbot-dns-do",
-		"version": "~=0.31.0",
+		"package_name": "certbot-dns-domainoffensive",
+		"version": "~=2.0.0",
 		"dependencies": "",
 		"credentials": "dns_do_api_token = YOUR_DO_DE_AUTH_TOKEN",
-		"full_plugin_name": "dns-do"
+		"full_plugin_name": "dns-domainoffensive"
 	},
 	"domeneshop": {
 		"name": "Domeneshop",


jc21:
@FabianK3 are you able to use the docker image mentioned to conduct your tests?

jc21:
Can I ask then, why do _you_ want to update this dns plugin when you don't use it?

FabianK3:
@jc21 Perhaps i didn't put it clearly enough!
I am planning on using this provider as i am owning domains on said provider.
I'll use the provider within NPM in my home lab, but i currently have not the time to get going with it.

As for why i updated the provider: I am sure the current provider implementation does what it is supposed to do. From a conversation with the do.de provider support i was informed that they do provide an API and linked me to their certbot implementation. I then realized it was a new fork in their name of the current provider used in NPM and my thought was it may be more future-proof to use their own fork instead of the current should any changes arise in the future.

Tl;dr: I have relations to do.de and i like to give back to this project. :)

(Merry Christmas!)

jc21:
Ok great I'll wait for you (or someone) to test it then.

If you would like CI to rebuild the PR docker image for you, simply leave a comment here with _only_ this text in the body:
```
rebuild
```

jc21:
I guess no one wants to test it. I'll release it and wait for complaints.

chindocaine:
I found this PR today because it breaks renewal for all existing certificates using this provider.
`Failed to renew certificate npm-* with error: The requested dns-do plugin does not appear to be installed`
Here's what I had to do to fix this:
```bash
sed -i 's/dns-do$/dns-domainoffensive/g' /etc/letsencrypt/renewal/*.conf
sed -i 's/dns_do_credentials/dns_domainoffensive_credentials/g' /etc/letsencrypt/renewal/*.conf
sed -i 's/dns_do_api_token/dns_domainoffensive_api_token/g' /etc/letsencrypt/credentials/credentials-*
```
Also I assume it won't even work for new ones since the `dns_do_api_token` was not updated.
This PR brings no value and also breaks existing logic, so I vote to revert it. @jc21 @FabianK3

FabianK3:
> I guess no one wants to test it. I'll release it and wait for complaints.

Perhaps another approach would've been wiser. 😁 
I feared this might happen without any testing.
@chindocaine @jc21 I agree to revert until it has been properly tested.

FabianK3:
Perhaps instead of reverting it, we could just add the fix from @chindocaine ?
I am not really aware on how to migrate content like that within a PR/release, any help would be appreciated.

chindocaine:
I also don't know if it would be possible to run the above mentioned fixes for the renewal in an automated way. I'm also not quite sure what would be the better way to go, revert the change (some people, including me, might have modified their config already, so for them it would break again) or keep it and hope that everyone who is affected by it will find this thread to know how to fix it.

Anyway, if we keep the new plugin, at least the config should be fixed to use
`"credentials": "dns_domainoffensive_api_token = YOUR_DO_DE_AUTH_TOKEN",`
instead of
`"credentials": "dns_do_api_token = YOUR_DO_DE_AUTH_TOKEN",`
I tested it in the mean time, without this fix the creation of a new certificate will fail as well. I guess I could make a PR for it.

Issue #4221, opened 2024-12-10
SSL Certificates show if cert is in use on host (and if host is disabled or enabled)


svefro:
On "SSL Certificates" page. please show if cert is in use on host (and if host is disabled or enabled)

**Is your feature request related to a problem? Please describe.**

**Describe the solution you'd like**
Add column with hosts that use the cert.

**Describe alternatives you've considered**

**Additional context**

PR #4244, opened 2024-12-18
SSL Certificates: Show if cert is in use on host (and if host is disabled or enabled)

badkeyy:
### Issue
Fixes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4221

### Description:
This PR enhances the SSL Certificates list by adding a Status indicator, indicating if a SSL Certificate is in use on a proxy host or not. If it is in use the domain names for which it is in use on are shown in the dropdown menu of the list item.

### Screenshots
image
image


badkeyy:
Sorry for setting the status of this PR back to Draft but I noticed that I forgot to implement this feature for dead_hosts and redirection_hosts

PR #4275, opened 2025-01-05
Force SSL Options Prevents Let's Encrypt Renewals

MeCJay12:
The force-ssl.conf file does the following:
1. Checks if the request is HTTP and sets $test to H if so
2. Checks if the request is to "test-challenge" in the normal Let's Encrypt challenge directory and appends T to $test if so
3. If $test = H (and only H) upgrades the request to HTTPS

This works but only for "test-challenge". All other Let's Encrypt renewals/challenges will fall through and be upgraded to HTTPS which causes them to fail (not sure why, see thoughts below). By modifying step two to catch all Let's Encrypt challenges and not upgrade those connections, renewals succeed.

It looks like Let's Encrypt should work without the modification since letsencrypt-acme-challenge.conf is included in the server with both the listen 80 and listen 443 directives. Either way, this has never worked for me. I've already implemented this change in my environment by mounting the modified force-ssl.conf file directly and it fixed the issue immedeatly.

### DIFF
--- a/docker/rootfs/etc/nginx/conf.d/include/force-ssl.conf
+++ b/docker/rootfs/etc/nginx/conf.d/include/force-ssl.conf
@@ -2,7 +2,7 @@ set $test "";
 if ($scheme = "http") {
 	set $test "H";
 }
-if ($request_uri = /.well-known/acme-challenge/test-challenge) {
+if ($request_uri ~* /.well-known/acme-challenge/) {
 	set $test "${test}T";
 }
 if ($test = H) {


jc21:
This acme challenge thing has always been a thorn in my side.

1. When a new host/certificate is being created, this should go to NPMs file structure
2. The community wants to pass through the letsencrypt challenge to their upstream services that may also use it
3. When the NPM renewal process is happening, the codebase doesn't know which host _might_ get renewed so it doesn't attempt to hijack that path

Point 1 is fine and works as expected, point 2 also AFAIK (I don't do this myself). But your issue is with point 3. It's hard to respect point 2 at the same time.

The backend code fires off the certbot renewal process, so it could put a temporary hijack in place during that execution.

MeCJay12:
This change should only affect Let's Encrypt flows by preventing them from being redirected to HTTPS. I don't think that this will impact anyone given that LE challenges on HTTP port 80 is standard. This should not impact the host routing of the requests; only the port & protocol.

That said, I do also have another fix for host routing of LE challenges to upstream hosts. Let me know if you are interested and I can open another pull request to discuss that in.

Issue #4276, opened 2025-01-05
MFA for Nginx Proxy Manager


Caruso8677:
I am submitting this feature request to see if in the near future there would be a way to have sometype of MFA for my account to login to Nginx Proxy Manager?

SirWobbyTheFirst:
+1 for this. In addition, LDAP or SSO authentication would also be a plus, I'd like to have Authentik provide the login UI for NPM instead of having to manage seperate users for NPM.

madtempest:
> +1 for this. In addition, LDAP or SSO authentication would also be a plus, I'd like to have Authentik provide the login UI for NPM instead of having to manage seperate users for NPM.

There is already a branch for open ID connect available. I don't know the details on when we can expect it to be merged, but OIDC for SSO support does exist.

badkeyy:
I will have a look at this soon

badkeyy:
Hey, sorry @jc21 it seems like I linked the wrong issue with the https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4283 PR. This issue is not yet to be closed. PR https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4283 closes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4221.

PR #4283, opened 2025-01-10
SSL Certificates: Show if cert is in use on host

badkeyy:
### Issue:
Fixes: https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4276

### Description:
This PR enhances the SSL Certificates list by adding a Status indicator, indicating if a SSL Certificate is in use on a proxy host or not. If it is in use the domain names for which it is in use on are shown in the dropdown menu of the list item.

### Screenshots
image
image


badkeyy:
Sorry for the duplicate PR but I had to recreate the PR because the Jenkins pipeline would not run anymore on the old one

PR #4334, opened 2025-02-01
Update certbot-dns-plugins.json (mijn-host)

nwagenmakers:
Updated credentials hint/text in mijn-host plugin entry


nwagenmakers:
Verified, looking okay for me!

PR #4338, opened 2025-02-03
Fix type for token.expires

Sander0542:
Currently, the `token.expires` property is advertised as an interger in the api schema, but the API returns a ISO string. I updated the schema to match the actual API.

https://github.com/NginxProxyManager/nginx-proxy-manager/blob/b4f49969d64c673acaa2f2a672f934ac84a66c5d/backend/internal/token.js#L70

https://github.com/NginxProxyManager/nginx-proxy-manager/blob/b4f49969d64c673acaa2f2a672f934ac84a66c5d/backend/internal/token.js#L131

https://github.com/NginxProxyManager/nginx-proxy-manager/blob/b4f49969d64c673acaa2f2a672f934ac84a66c5d/backend/internal/token.js#L159

### DIFF
--- a/backend/schema/components/token-object.json
+++ b/backend/schema/components/token-object.json
@@ -5,10 +5,9 @@
 	"additionalProperties": false,
 	"properties": {
 		"expires": {
-			"description": "Token Expiry Unix Time",
-			"example": 1566540249,
-			"minimum": 1,
-			"type": "number"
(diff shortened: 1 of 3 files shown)


jc21:
Hmm right you are. It used to be a UNIX timestamp integer, not sure when that changed but the frontend doesn't really use it anyway.

PR #4358, opened 2025-02-09
use latest certbot-dns-cpanel version

pustekuchen91:
this allows to use token for authentication

Because an old version of the `certbot-dns-cpanel` is used, its now possible to use tokens for authentication

https://github.com/NginxProxyManager/nginx-proxy-manager/discussions/3057#discussioncomment-10805031


pustekuchen91:
hey @jc21, 
who could review / merge ?

PR #4406, opened 2025-02-28
Fix DomainOffensive certbot plugin

chindocaine:
In https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4235 the certbot plugin for do.de (Domain Offensive) was updated to use the more official version. One necessary line modification was missed, resulting in an error when creating a new certificate.

Also see the comments on the above mentioned PR in case you came here because cert renewal is broken.

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -164,7 +164,7 @@
 		"package_name": "certbot-dns-domainoffensive",
 		"version": "~=2.0.0",
 		"dependencies": "",
-		"credentials": "dns_do_api_token = YOUR_DO_DE_AUTH_TOKEN",
+		"credentials": "dns_domainoffensive_api_token = YOUR_DO_DE_AUTH_TOKEN",
 		"full_plugin_name": "dns-domainoffensive"
 	},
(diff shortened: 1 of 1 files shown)


FabianK3:
Thank you for the fix.

Like yourself others had the same issue regarding not-transitioned values, is changing the variable name enough for transitioning everything else?

You mentioned that you renamed a bit more when you encountered the issue (https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4235#issuecomment-2677066481).

chindocaine:
Exactly, the linked comment describes what I had to to to get my existing certificates to renew again. Unfortunately I don't know enough about the inner workings of the Nginx Proxy Manager to figure out if this could be automated for other users, so this PR doesn't address this issue.
It rather fixes the creation of new certificates with the do.de provider, which is currently also broken. In that case the problem is just the variable name that wasn't updated.

FabianK3:
> And tbh I'm not sure if it's still worth fixing, because it has been broken for about 2 months now, so most existing certs will have expired anyway.

Valid point, i think that is acceptable.

Thank you for that fix!

PR #4422, opened 2025-03-07
fixed what seems to be a typo

dvdpearson:
While trying to setup the dev environment, I realized the docker-compose file was trying to mount a file that doesn't exist.

Was it missed somewhere?


jc21:
It's not a typo. the tmp file is created by the start dev script, which replaces strings in the template for dynamic container Ip addresses.

PR #4512, opened 2025-04-28
query by domain endpoint

code-a-cola:
We have automated script that run outside of NPM. These scripts use the backend api to add and remove proxies, as well as query for existing proxies.

We have scripts the need to pull specific proxies by domain, rather than proxy_id.

Pulling the entire list of proxies to search for a proxy by domain is extremely slow, especially if a large amount of proxies exist.

This pull request will add an endpoint to the backend API to allow the query of proxies by domain name.


jc21:
For some reason, CI doesn't want to build this PR. Perhaps rebasing it in your fork might help?

PR #4557, opened 2025-05-26
fix(modal): make textarea font actually monospace

1ukastesar:
Modal `textarea` element has this class `text-monospace`, but there is actually no CSS definition that sets the monospace font for it (neither in custom SCSS files, nor in included libs). This commit fixes the issue by setting `monospace` `font-family` for the `textarea`, greatly enhancing UX of configuration editing in UI.


1ukastesar:
@jc21 can you please merge this PR so I don't have to stay on my branch with this fix?

Issue #4606, opened 2025-07-01
Cannot Log into Admin After Upgrade to 2.12.4


paelan8a:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes ([#1769](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/1769))

**Describe the bug**
After upgrading from 2.12.3 to 2.12.4, I can no longer log into the admin panel. The login page is displayed, but authentication does not work. The Docker logs show the following and do not proceed any further:
```
nginx  | ❯ Starting backend ...
nginx  | [7/1/2025] [11:29:07 AM] [Global   ] › ℹ  info      Using Sqlite: /data/database.sqlite
nginx  | [7/1/2025] [11:29:08 AM] [Migrate  ] › ℹ  info      Current database version: none
nginx  | [7/1/2025] [11:29:08 AM] [Global   ] › ⬤  debug     CMD: [ -f '/etc/letsencrypt/credentials/credentials-50' ] || { mkdir -p /etc/letsencrypt/credentials 2> /dev/null; echo '# Cloudflare API token
nginx  | dns_cloudflare_api_token=****' > '/etc/letsencrypt/credentials/credentials-50' && chmod 600 '/etc/letsencrypt/credentials/credentials-50'; }
nginx  | [7/1/2025] [11:29:08 AM] [Global   ] › ⬤  debug     CMD: [ -f '/etc/letsencrypt/credentials/credentials-63' ] || { mkdir -p /etc/letsencrypt/credentials 2> /dev/null; echo '# Cloudflare API token
nginx  | dns_cloudflare_api_token=****' > '/etc/letsencrypt/credentials/credentials-63' && chmod 600 '/etc/letsencrypt/credentials/credentials-63'; }
nginx  | [7/1/2025] [11:29:08 AM] [Certbot  ] › ▶  start     Installing cloudflare...
[...]

PR #4607, opened 2025-07-01
Update /frontend/html/login.ejs

tom-kst:
**Bug found:** `/js/login_bundle.js` has been changed to `/js/login.js`
------
This caused dashboard page could be accessed, but can't login after input ID & PWD...


Joshndroid:
is this a fix for #4606 ?

paelan8a:
It is also frontend bugfix, so #4606 is a different issue.

tom-kst:
I’m sorry, it’s not the file names or path issue, shame on me 🤪🤪

PR #4615, opened 2025-07-01
Revert certbot-dns-plugins.json changes to credentials placeholder

picnicbob:
Storing user/password instead of an API token is not a good idea and specifically not recommended by [certbot-dns-cloudflare documentation](https://certbot-dns-cloudflare.readthedocs.io/en/stable/index.html#certbot-cloudflare-token-ini).
Spaces were intentionally removed in the past by #3781.
`dns_cloudflare_api_token=abc123` was/is working fine for me.

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -68,7 +68,7 @@
 		"package_name": "certbot-dns-cloudflare",
 		"version": "=={{certbot-version}}",
 		"dependencies": "cloudflare==4.0.* acme=={{certbot-version}}",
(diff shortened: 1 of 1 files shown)


mcpattrsn (approved):
Documentation reference is correct and works with dependency corrections. Maybe add restricted and global options?

jc21:
Plugins file has moved [here](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/backend/certbot/dns-plugins.json#L70) and already has this change.

PR #4638, opened 2025-07-06
[Dockerfile] remove unnecessary packages.

behnejad:
removing unnecessary packages to minimize the docker image.


jumbi77:
@jc21 This seems right and easy. Any objections to merge this?

jc21:
There was a reason it was added in the past.. but in general it should be fine, just need to ensure it won't affect node somehow.

behnejad:
👍

PR #4660, opened 2025-07-16
Proxy protocol Support

NikeLaosClericus:
With Iaotle fix


NikeLaosClericus:
This was mostly just @adrum's pr #4505 with a merged fix from @Iaotle so it would build for testing.
@adrum has since superseded this in their original pr.

PR #4739, opened 2025-08-26
Feature: Load balancer functionality

Jacob-Tate:
This was a quickly put together implementation of NPM with the ability to add and manage upstreams. Im very open to comments since i just kinda hacked this together so I could use this for myself ive never worked in most of these frameworks so if anything looks wrong let me know :)

This adds an upstream tab where you can manage the upstreams avaliable maybe this should be put somewhere else im not sure this worked for now! You then can select the upstream in the proxy_host see screenshots below. At the moment if you want a weight you have to set weight=2 instead of just 2 since there are a few other options there that i actually use like turning off hosts and whatnot

image
image
image
image
image
image

### DIFF
--- a/backend/internal/proxy-host.js
+++ b/backend/internal/proxy-host.js
@@ -49,6 +49,15 @@ const internalProxyHost = {
 				data.owner_user_id = access.token.getUserId(1);
 				data               = internalHost.cleanSslHstsData(data);
 
+				// If upstream is used, clear forwarding fields
+				if (data.upstream_id) {
+						data.forward_host = '';
+						data.forward_port = 0;
+				}
+
+				// This is a UI-only field, remove it
+				delete data.forward_to_type;
+
 				// Fix for db field not having a default value
 				// for this optional field.
 				if (typeof data.advanced_config === 'undefined') {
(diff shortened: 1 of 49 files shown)


Jacob-Tate:
Now that im thinking of it maybe putting the upstream form within the forward proxy form would be better instead of separating them does anyone have any ideas if thats a better idea?

id look something like this (photoshopd)
Untitled

Heroj04:
I think it may still be worth it to have them separated as you could want to include the same upstream for proxy hosts and stream hosts if possible?

Jacob-Tate:
> I think it may still be worth it to have them separated as you could want to include the same upstream for proxy hosts and stream hosts if possible?

Thats actually a great point I dont use the streams ill setup a test and implement that.

tqtuan1512:
How can I install it ? Please give me step guide. thanks so much

jc21:
Great work! But I have some bad news, I rewrote the interface in React and merged it to develop.. Your changes could still be applied, but it will need more work and thought.

I was working on upstreams in the v3 branch, but I'm moving some of that functionality back to v2, and this work is great.

Issue #4692, opened 2025-08-05
Setting `scheme: auto` via API or UI results in invalid redirect behavior


sandeepgs24x7:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes ( part of the issue being reported here is already reported in this ticket - https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4080 )

**Describe the bug**
When creating a redirect host using the API endpoint `api/nginx/redirection-hosts` with `"forward_scheme": "auto"`, the redirect results in a broken URL like `auto://google.com`. This is not a valid scheme and causes the redirect to fail.

Additionally, when attempting to set `"auto"` as the scheme in the web UI, the following validation error is shown: `data/forward_scheme must be equal to one of the allowed values`  

This reveals inconsistent behavior:
- The **API** accepts `"auto"` but results in a broken redirect.
- The **UI** correctly rejects `"auto"` as invalid although it is an option in the form.

**Nginx Proxy Manager Version**
v2.12.6

**To Reproduce**
**- Using the REST API:**
1. Send a `POST` request to `/api/nginx/redirection-hosts` with the following JSON payload:

```
   {
     "domain_names": ["mydomain.com"],
     "forward_scheme": "auto",
     "forward_domain_name": "google.com",
     "forward_http_code": 300,
     "preserve_path": true
   } 
``` 

2. Access http://mydomain.com in a browser.
3. Observe that the redirect attempts to go to auto://google.com.

**- Using the UI:**
1. Open the Nginx Proxy Manager web UI.
2. Go to the "Redirect Hosts" section and click "Add Redirection Host".
3. Fill in domain and forward details.
4. Set Scheme = auto.
5. Try to save.
6. See the following error: `data/forward_scheme must be equal to one of the allowed values`  

**Expected behavior**
1. If "auto" is intended to be a supported scheme, the redirect should dynamically use the original request's scheme (http or https).
2. If "auto" is not a supported value, it should be rejected consistently across both the API and UI with a validation error.

**Additional context**
A related issue ([#4080](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4080)) covers the UI validation problem, but this report also includes the API behavior and resulting redirect failure.

Spacelord09:
Related PR: #4751

Issue #4080, opened 2024-10-17
Unable to setup any Redirection Host


twyxxywt:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
On version 2.12.1 (which I updated to yesterday), seem to rendered setting up any new Redirection Host broken.

No matter what domain I put in in "Domain Names" or in "Forward Domain". Kept getting the message "data/forward_scheme must be equal to one of the allowed values".

**Nginx Proxy Manager Version**
v2.12.1

**To Reproduce**
1. Go to 'Redirection Hosts'
2. Click on 'Add Redirection Host'
3. Add any new "Domain Names" and "Forward Domain"
4. Error in red pops up - "data/forward_scheme must be equal to one of the allowed values"

**Expected behavior**
New Redirection Host get added without any errors.

**Screenshots**
<img width="530" alt="image" src="https://github.com/user-attachments/assets/a8fe5a52-7890-471d-a2ed-4d51453560f2">

**Operating System**
Hosted in Debian 12. Standard Docker install.

**Additional context**
Tried it on multiple NPM instances running 2.12.1. Ran on a NPM running version "v2.11.3" and it's working fine.

SmokeDevil666:
I have the same problem

madushanj:
Same for me on version 2.12.1

haye86:
Same for me also version 2.12.1 
also existing hosts cannot be altered, same error. 
seems like SSL certs for the redirection hosts cannot be renewed due to this error.

9021007:
Can confirm v2.12.1

ki4arl:
Can confirm v2.12.1

danfratamico:
Having the same issue

omidbaharifar:
Same issue
in my case, setting "Scheme" field to "http" or "https" fixes the problem. (It seems like there is a problem with processing the "auto" value for Scheme in the backend).

9021007:
Sure, though the issue is now that it now forces you to pick http or https, instead of just keeping whatever you want.

trulow:
I'm also encountering this issue on the lastest docker version of 2.12.1

Huythanh0x:
And resolved the same problem by changing the `Scheme` option to `http` from `auto`.
[...]

PR #4751, opened 2025-08-31
Fix Bug in Redirection Host Schema preventing auto redirection hosts

cal-tec:
This commit fixes bug's #4692 and #4080  A previous attempt to fix this issue ( commit d96a398 ) referenced the wrong string from the ui.

### DIFF
--- a/backend/schema/components/redirection-host-object.json
+++ b/backend/schema/components/redirection-host-object.json
@@ -28,7 +28,7 @@
 		},
 		"forward_scheme": {
 			"type": "string",
-			"enum": ["auto", "http", "https"]
+			"enum": ["$scheme", "http", "https"]
 		},
 		"forward_domain_name": {
 			"description": "Domain Name",


Spacelord09:
Can anyone merge this?

ebirbe:
Hello @jc21 are you the right person to review / merge this?

spotter-mmf:
I'm running into this issue with redirection hosts as well. Hoping this will fix the problem, if we can get the change merged?

jc21:
Can you test with the new react UI:
```
nginxproxymanager/nginx-proxy-manager-dev:develop
```
and let me know if it's still an issue?

spotter-mmf:
Works without an issue for me in that build!

Spacelord09:
#4692

jc21:
I've fixed this differently, by using `auto` for the UI and db column and rendering `$scheme` in the nginx config.

The reason for this: I don't like the idea of having this value being set in the nginx config directly from database data without any validation. I also don't like using dollar variables _as_ values.

jc21:
Thanks for your contribution though! I do appreciate the effort!

PR #4762, opened 2025-09-12
Add support for custom certificate key types via environment variables

mz2:
Adds support for custom certificate key types through `CERT_KEY_TYPE` and `CERT_ELLIPTIC_CURVE` environment variables, optionally also passed to the Dockerfile, to enables ECDSA P-256 certificates and other key types beyond the default.

Maintains backward compatibility when environment variables are empty or not set (i.e. no impact to current behaviour when the vars are not set).

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -857,6 +857,13 @@ const internalCertificate = {
 			certificate.domain_names.join(','),
 		];
 
+		if (process.env.CERT_KEY_TYPE) {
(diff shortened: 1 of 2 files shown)


jc21:
Looks good, could you please add a section to the documentation related to this:

https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/docs/src/advanced-config/index.md

So that your changes are not hidden :)

PR #4770, opened 2025-09-19
Fix certificates shown as Inactive when used for redirection

### LINKED ISSUE #4455: Certificate show as inactive if is only used for redirection
**Checklist**
  - Yes
  - Yes
  - Yes

**Describe the bug**

- When created a new certificate if is only used for a redirection show as inactive, this happened after #4283 
- Another problem is the PR only show hosts, not redirection.
- Another not tested thing is this bug will happens if only stream is used with the certificate

2.12.3

**To Reproduce**
1. Create a certificate
2. Assign to a redirection only
3. Certificate will keep as inactive

**Expected behavior**
Show certificate as used if assigned to host, redirection or stream

**Screenshots**

Image

Image

Giotino:
`frontend/js/app/nginx/certificates/list/item.js` contains a typo: `redirect_hosts` instead of `redirection_hosts`.

This causes the STATUS column on the "SSL Certificates" page to break, as redirect hosts are not detected and are therefore shown as Inactive.

See #4455 for more information.

Fixes #4455

### DIFF
--- a/frontend/js/app/nginx/certificates/list/item.js
+++ b/frontend/js/app/nginx/certificates/list/item.js
@@ -52,8 +52,8 @@ module.exports = Mn.View.extend({
             },
             dns_providers: dns_providers,
             active_domain_names: function () {
-                const { proxy_hosts = [], redirect_hosts = [], dead_hosts = [] } = this;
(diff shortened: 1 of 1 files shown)


jc21:
This PR should be resolved with changes in the new react version that is now merged. Please reopen if you still find problems.

PR #4771, opened 2025-09-19
Better exploit blocking

fuomag9:
This should improve exploit blocking, better implementation would to be use https://coraza.io but in the short term this should be more useful. I've removed the spam section since it doesn't make sense and it's too prone to bypasses anyway (I know obfuscation also works for other vulnerabilities types)

### DIFF
--- a/docker/rootfs/etc/nginx/conf.d/include/block-exploits.conf
+++ b/docker/rootfs/etc/nginx/conf.d/include/block-exploits.conf
@@ -1,6 +1,7 @@
 ## Block SQL injections
 set $block_sql_injections 0;
 
+# Traditional SQL injection patterns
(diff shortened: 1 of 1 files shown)


jc21:
Sorry I'm not going to merge a PR that changes 616 files that are not specific to the subject of the PR.

I don't know what you're trying to accomplish with these changes, but it looks like you're rewriting the entire project.

fuomag9:
> Sorry I'm not going to merge a PR that changes 616 files that are not specific to the subject of the PR.
> 
> I don't know what you're trying to accomplish with these changes, but it looks like you're rewriting the entire project.

You can disregard this, since the PR was intended only for the commit at the time of writing (and I forgot about this pull request)

PR #4148, opened 2024-11-07
Added certbot plugin for Beget DNS service

Medan-rfz:
Added support for DNS provider Beget using the sertbot plugin certbot-dns-beget

PR #4772, opened 2025-09-22
Change Beget plugin

ape364:
Hello.

[Current plugin](https://pypi.org/project/certbot-beget-plugin/) is outdated and not working (added by https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4148)
Changed to working one: https://pypi.org/project/certbot-dns-beget-api/

Maybe there is any chance to edit plugin name and use it before merging & update?

### DIFF
--- a/global/certbot-dns-plugins.json
+++ b/global/certbot-dns-plugins.json
@@ -41,11 +41,11 @@
 	},
 	"beget": {
 		"name":"Beget",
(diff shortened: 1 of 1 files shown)


jc21:
The plugins json file has moved [here](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/backend/certbot/dns-plugins.json), please apply your change there.

PR #4784, opened 2025-10-01
Fixed disappearing domain names when focus is removed from field

VicentCano:
What you have written in domain names fields (Adding hosts, certificates, etc.) disappears when you move to any other field without clicking or pressing enter. This is really inconvenient and counterintuitive. This PR fixes this.


jc21:
This PR no longer applies now that the React rewrite is merged. Thanks for the contribution, but I'll close this PR. If you still find the new React UI problematic, please open a new PR.

PR #4794, opened 2025-10-08
Add basic MySQL TLS support

Johno-ACSLive:
### Summary

This PR adds **minimal, opt‑in TLS support** for MySQL connections in Nginx Proxy Manager. It introduces three environment variables:

| Variable | Default | Purpose |
|----------|---------|---------|
| `DB_MYSQL_SSL` | `false` | Enable SSL when set to true. |
| `DB_MYSQL_SSL_REJECT_UNAUTHORIZED` | `true` | Controls certificate chain validation. |
| `DB_MYSQL_SSL_VERIFY_IDENTITY` | `true` | Enables hostname verification. |

No behavior changes for existing users unless `DB_MYSQL_SSL` is explicitly enabled. Does not impact other DB engine configuration.

### Motivation

Environments such as:
- Azure Database for MySQL (secure transport enforced)
- Self-hosted MySQL with `require_secure_transport=ON`
- Security baselines requiring encrypted DB connections

…currently cause startup failure (`ER_SECURE_TRANSPORT_REQUIRED`) because connection attempts are plaintext. This update allows secure connections while remaining completely backward compatible.

### Backward Compatibility

| Scenario | Before | After |
|----------|--------|-------|
| No TLS vars set | Plain MySQL | Same |
| Server requires TLS; no vars set | Startup error | Same (user must enable SSL) |
| TLS enabled (`DB_MYSQL_SSL=true`) | N/A | Encrypted session |
| Self-signed cert (user sets `DB_MYSQL_SSL_REJECT_UNAUTHORIZED=false`) | N/A | Connects (user-accepted risk) |
| Hostname mismatch & `DB_MYSQL_SSL_VERIFY_IDENTITY=true` | N/A | Fails (correct security posture) |

### Out of Scope / Future Work

- Optional CA / client certificate path variables
- Min/Max Cipher and associated checks
- Postgres TLS parity (could mirror same pattern)


jc21:
Hi, please rebase and hopefully things will pass now.

Johno-ACSLive:
> Hi, please rebase and hopefully things will pass now.

Hi, done :)

PR #4812, opened 2025-10-27
Remove duplicate domains on Certificates v.2.13.0

sopex:
* Removes duplicate domains
* Testing blue as a potential colour with better contrast


sopex:
No need for it anymore.

Issue #4802, opened 2025-10-19
Hetzner "DNS Console is moving to the Hetzner Console"


tsohst:
**What provider would you like to see added to NPM?**
Hetzner is moving DNS Console to the Hetzner Console.
After migration in Hetzner to Hetzner Console, DNS challenge seems broken because of "zone not found".

```
[10/19/2025] [7:27:03 PM] [Global   ] › ⬤  debug     CMD: certbot certonly --config '/etc/letsencrypt.ini' --work-dir "/tmp/letsencrypt-lib" --logs-dir "/tmp/letsencrypt-log" --cert-name 'npm-32' --agree-tos --email 'XXXX' --domains 'XXXX' --authenticator 'dns-hetzner' --dns-hetzner-credentials '/etc/letsencrypt/credentials/credentials-32' 

[10/19/2025] [7:27:06 PM] [Global   ] › ⬤  debug     CMD: /usr/sbin/nginx -t -g "error_log off;"

[10/19/2025] [7:27:06 PM] [Nginx    ] › ℹ  info      Reloading Nginx

[10/19/2025] [7:27:06 PM] [Global   ] › ⬤  debug     CMD: /usr/sbin/nginx -s reload

[10/19/2025] [7:27:06 PM] [Express  ] › ⚠  warning   Saving debug log to /tmp/letsencrypt-log/letsencrypt.log

Encountered exception during recovery: certbot_dns_hetzner.hetzner_client._ZoneNotFoundException: Zone port.movtube.de not found in Hetzner account

Zone port.movtube.de not found in Hetzner account
```

Before:
```
curl -sS https://dns.hetzner.com/api/v1/zones   -H "Auth-API-Token: XXXXX"
{"zones":[],"meta":{"pagination":{"page":0,"per_page":0,"previous_page":0,"next_page":0,"last_page":0,"total_entries":0}},"error":{"message":"zone not found","code":404}}

```
After the migration
```
        curl -H "Authorization: Bearer XXXXX" \
        "https://api.hetzner.cloud/v1/zones"
{
    "meta": {
        "pagination": {
            "last_page": 1,
            "next_page": null,
            "page": 1,
            "per_page": 25,
            "previous_page": null,
            "total_entries": 1
        }
    },
    "zones": [
        {
```

lastsamurai26:
Please read this 
https://docs.hetzner.com/de/networking/dns/migration-to-hetzner-console/process/

[...]

PR #4813, opened 2025-10-27
Adds Hetzner Cloud to available plugins

potatojuicemachine:
This MR adds Hetzner Cloud as a certbot dns provider because hetzner is moving from dns console to cloud.  
Closes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4802


potatojuicemachine:
I fail to understand why my PR can't be built. Can anyone help me with this?  
Additionally I am unable to get the dev environment up and running as the scripts always fail with an error message related do postgres and unrelated (afaik) to my changes.  
It as been multiple hours of failing and any help would be appreciated.

renanmedina:
@potatojuicemachine I also got the postgres error, its because the setup is using :latest tag of postgres which is always not advisable. (seems the latest tag does some new checks and dont get the container up and running)

I fixed that by just commenting out the connection in the docker-compose.dev file to use the sqlite version.

image

that way it ignores the postgres container and let you work normally. 

I'm not sure how the ci in jenkins is implemented currently, but if it's also using postgres in there, it should be the reason why its failing for everyone (it failed to me in my PR)

potatojuicemachine:
@renanmedina I think it could be this issue. At least I had the same with postgres in a different context https://forums.docker.com/t/mysterious-docker-issue-with-postgres/149898/3

potatojuicemachine:
I tested the built container and I was able to get a certificate from hetzner cloud.

sopex:
Maybe PR it against the react branch, since 2.13 is based on that, and it's close to going live.

potatojuicemachine:
@sopex I am not sure what the correct way to contribute is :) I changed the base

sopex:
@jc21 Can we test this in 2.13? Thank you a lot!

Issue #257, opened 2019-12-15
Custom Timeout


JonathanTreffler:
The default timeout of Proxy Hosts seems to be 1 Minute, wich is pretty short, so i would like to change the timeout without having to use a Custom Nginx Configuration.

**Describe the solution you'd like**
A input field in the Web Interface to change the timeout per Proxy Host.
It should probably set following nginx configuration parameters:
- proxy_connect_timeout
- proxy_send_timeout
- proxy_read_timeout
- send_timeout

**Additional context**
![image](https://user-images.githubusercontent.com/28999431/70865961-bd842980-1f63-11ea-9c55-3bb4104a63a4.png)

danmed:
Def need this.. so many things i reverse proxy need a much longer timeout.. Anything to do with Docker usually takes more than a minute to build a container and NPM causes it to bail out.

TessyPowder:
Maybe i will do a PR for this Feature at some point, but i currently don't have enough time

Bloham:
I really need that 👍  Hope you get around to it sometime!

mostafabarmshory:
+1

c-schulz-rd13:
Would like to see that too! Good feature request!

mangopudding:
+1.

Is there a workaround in the meantime?

marcwimmer:
+1

joelloube:
+1

Has anyone figured out a workaround?

Hikari:
+1

tryanbot:
no update regarding this?

chaptergy:
The "workaround" is: just add the parameter you want to change to the advanced config field. If you want to change the proxy timeout, set the corresponding proxy timeout variables. 

I am using the following line of advanced custom config to prevent console disconnects in Portainer:

```nginx
proxy_read_timeout 30m;
```

tryanbot:
super helpful, thank you so much.

chaptergy:
@mriksman It worked for me as well, with timeouts in the portainer console. Maybe whatever app you proxy has some internal timeouts which expire independent of whether the proxy connection is kept open?

Oh and @JonathanTreffler, why is it exactly that you want this feature without the custom nginx configuration? That's exactly what the custom config is for.

Padillam5:
> 

did you ever find a solution to cloudflare timeout

hoangnamitc:
[...]

PR #4815, opened 2025-10-27
UI - Custom timeouts setup

renanmedina:
Related to: https://github.com/NginxProxyManager/nginx-proxy-manager/issues/257

image

I also needed this feature to make it faster to setup multiple proxies without the need for the custom configuration in all of them.

I heavily use Nginx Proxy Manager, and even though creating custom rules are easy enough, having the option directly in the UI makes life easier and setups faster.

**Maybe the best way would actually to have a "Timeouts" tab instead of just putting it directly into advanced tab??**


jc21:
Now that the React rewrite is merged, this PR is very much outdated. It might be easier to create a new PR with your changes, as the backend API has also had some rewrites.

renanmedina:
> Now that the React rewrite is merged, this PR is very much outdated. It might be easier to create a new PR with your changes, as the backend API has also had some rewrites.

@jc21 when did this rewrite got merged? When i did this i got the most updated version from develop branch. 

just wanted to know the timeline there, I will rewrite and re-open this using React then 👍🏻 

Thank you!

PR #4848, opened 2025-11-05
Add Japanese language support and translations

Oka-Tak:
Add Japanese support/localization

Co-authored-by: kz2870 <kz2870@users.noreply.github.com>


jc21:
Thanks. I just added a new translation for `redirection-host.forward-http-code`, can you add that in? I guess it can be a copy of the value for `column.http-code`

PR #4850, opened 2025-11-06
Add Spanish language support and translations

TeenBiscuits:
Add Spanish localization


TeenBiscuits:
Emmmm, what?

vsc55:
Hi @TeenBiscuits,
It appears the system failed during the check.
Try creating a new pull request or a new commit to your repository in the same branch to see if the system runs the check again.

jc21:
Thanks. I just added a new translation for `redirection-host.forward-http-code`, can you add that in? I guess it can be a copy of the value for `column.http-code`

TeenBiscuits:
Added. It's not a copy of the other value. `column.http-code` translates to "Access" in English while `redirection-host.forward-http-code` is "HTTP Code". I think this is a mistake; I didn't notice it while translating to Spanish. I fixed it in my last commit

jc21:
> column.http-code translates to "Access" 

Yeah I fixed that one on the develop branch yesterday.

PR #4870, opened 2025-11-08
Add Russian Support

kraineff:


jc21:
thanks. I just added a new translation for `redirection-host.forward-http-code`, can you add that in? I guess it can be a copy of the value for `column.http-code`

kraineff:
Sure, I will add all missing strings in a few days!

PR #4894, opened 2025-11-09
Update _access.conf to fix access_list.pass_auth logic

eduardpaul:
Wrong logic to pass auth as header: when disabled (pass_auth=0) credentials are included in Authorization header. However as soon as you enable (pass_auth=1) they are not.


jc21:
Good catch :)

Issue #4855, opened 2025-11-07
Firefox layout offset caused by calc(100vw - 100%) in UI stylesheet


vsc55:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes
- Are you sure you're not using someone else's docker image?
  - Yes
- Have you searched for similar issues (both open and closed)?
  - Yes

**Describe the bug**
In Firefox, when a vertical scrollbar appears, the main content area shifts to the right, leaving an empty space on the left side of the screen.
This issue doesn’t occur in Chrome.
The cause seems to be the following CSS rule in the main stylesheet:

```css
@media (min-width: 992px) {
  :host, :root {
    margin-left: calc(100vw - 100%);
    margin-right: 0;
  }
}
```
Firefox includes the scrollbar width in 100vw, which results in a visible horizontal offset.
Commenting out the margin-left line fixes the issue without breaking layout in any tested browser.

**Nginx Proxy Manager Version**
2.13.1

**To Reproduce**
1. Open Nginx Proxy Manager UI in Firefox (any version).
2. Navigate to a page that triggers vertical scrolling.
3. Observe a left-side space appearing equal to the scrollbar width.

**Expected behavior**
The layout should remain aligned to the left with no extra space when the scrollbar is active, consistent with Chrome’s behavior.

**Screenshots**
Screenshot showing the layout shift in Firefox (with scrollbar active).
<img width="1923" height="160" alt="Image" src="https://github.com/user-attachments/assets/6cdb8763-8eca-4c6e-8276-14e53969c105" />

Screenshot showing correct layout after commenting out the margin-left rule.
<img width="1919" height="126" alt="Image" src="https://github.com/user-attachments/assets/e44f0f64-9089-4d5f-9405-b58c2fc371df" />

**Operating System**
Windows 10, Firefox 144.0.2.

lastsamurai26:
I also experienced this during the beta phase of 2.13, but I could only reproduce it in Chrome-based browsers.
Since Windows 10 is no longer officially supported by MS, it is also possible that programs will discontinue their support. 

Unfortunately, I cannot reproduce this in Firefox (Windows 11 and macOS).

jc21:
I'm using Brave, which is chrome based, and I don't see this problem.

That css snippet above comes from [Tabler](https://tabler.io/)

lastsamurai26:
[...]

PR #4907, opened 2025-11-11
v2.13.3

jc21:


lastsamurai26:
Did you have a fix for this ? 
https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4855

i have it only in Chrome based Browser

PR #4910, opened 2025-11-11
Add scheme back in destination

### LINKED ISSUE #4846: Destination with Scheme (and Link)
I would appreciate it if the scheme were displayed before "Destination,"
Image
as in the previous 2.12.x version.

Image

As it is, it's difficult to tell whether it's http or https without editing the entry.

It would also be nice if this were a link.

7heMech:
Closes #4846

Not sure if removal of destination was intentional, it honestly doesn't bother me since all my hosts are http, but if it's what you intended feel free to close this.


jc21:
Good catch, thanks :)

Issue #4912, opened 2025-11-11
Wrong title when adding a custom SSL Certificate


peterge-misoft:
When you click on "Add Certificate" > "Custom Certificate":
<img width="241" height="224" alt="Image" src="https://github.com/user-attachments/assets/76ba4a37-aa05-4ead-b156-51682a897b76" />

The title of the window has the wrong title in v2.13.3:
<img width="641" height="642" alt="Image" src="https://github.com/user-attachments/assets/7b24c7be-ee26-433e-97ec-843c4a153ad1" />

PR #4917, opened 2025-11-12
Fix: wrong translate and adding missing translations 

lastsamurai26:
@jc21
I tried to fix the error in https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4912 myself. 
Please take a look and see if I did it correctly. 

I also added two German keys that I had overlooked. 
For some reason, certificate.renew seems to have been missing from the German translation.


jc21:
Heh I literally fixed the `certificate.renew` missing german translation like 10 mins ago. 

The dialog title change looks good though

lastsamurai26:
Okay, if you've done it, then you can ignore them. I just saw it and thought I'd do it too.

I forked the branch before I checked it. It must have overlapped.

You'll have to check again :-)

lastsamurai26:
Then only the certificate display needs to be merged.

Please check again (:

PR #4926, opened 2025-11-12
Fix unaligned flag on main screen when language is selected

7heMech:
Fixes this issue on the main screen, might've been better if I applied styles there, idk.
before and after:
image
image


7heMech:
I'll be fixing this in another PR with other stuff on main screen.

PR #4930, opened 2025-11-12
Update German translations

blinkerfluessigkeit:


bogo22:
LGTM (as a German speaker)

vwccvr6:
Seems to be good. 

I would prefer a different translation for "  "login.title": "Melden Sie sich bei Ihrem Konto an","

It's way to long, i would prefer "Anmelden" or keep it the english way "Account Login". I think that is pretty clear as a German and "eingedeutscht"

blinkerfluessigkeit:
Yeah, it is a bit too long for my taste as well, but I don't particulary like "Account Login". We could go with "Anmelden" like you suggested, or alternatively with "Einloggen". I don't have strong feelings about this either way as I use the UI in English anyway, so you can just decide if you want :)

jumbi77:
> Yeah, it is a bit too long for my taste as well, but I don't particulary like "Account Login". We could go with "Anmelden" like you suggested, or alternatively with "Einloggen". I don't have strong feelings about this either way as I use the UI in English anyway, so you can just decide if you want :)

Lets do "Anmelden", ok? Rest is fine.

blinkerfluessigkeit:
Done

jc21:
I've been a bit behind, but if you still want this change can you please rebase and resolve conflicts?

jc21:
And can someone tell me if #4946 is conflicting with this?

blinkerfluessigkeit:
No worries, I rebased onto develop. Yes, #4946 does conflict with my PR, so it will have to be rebased after merging #4930

PR #4967, opened 2025-11-19
Dns resolution solution

savergiggio:
Containerized service that automatically solves dynamic IP issues in Nginx Proxy Manager. Periodically monitors configured domains, detects IP address changes, and updates Nginx configurations in real-time.


7heMech:
Just no. There's no place for this in NPM, you can integrate it via API or keep it as external project, but if these features were to be implemented in this project, they'd have to be done as changes to the existing JS backend, not adding an additional python dependency.

PR #4975, opened 2025-11-20
feat: Clone a Proxy Host

Tacioandrade:
Hello! I'm sharing a feature I needed to make the work of the N2 support teams at the companies I work with easier.

I have some clients with specific source blocking policies, among other things, that the client's N2 support team always ended up getting wrong when configuring. For this reason, I developed this feature.

On the Proxy Hosts screen, in addition to Edit, Disable, and Delete, there's now a Clone button. When pressed, it opens the same Add a Proxy Host modal, but with the original Proxy Host data already filled in. You only need to change the host, IP, and port, and a new Reverse Proxy with the necessary settings is configured.

I would greatly appreciate it if you could add this feature to the official project so that you don't have to apply this patch with each new npm version.


Tacioandrade:
That's a strange error. I had managed to run it using the project's start-dev script.

However, I will fix this error and try the PR again by the end of the week. =)

Tacioandrade:
I understand the problem; it was just the way I wrote a comparison that Yarn flagged as non-standard.

I adjusted it and it seems to be working. =)

Tacioandrade:
Yes, I'd like to add that! This functionality is already implemented in npmplus, but I'd like to add it to npm as well.

If you could tell me what will be required for this PR to be accepted, I would appreciate it.

PR #4992, opened 2025-11-25
Handle postgres ssl

gsalingu:


jc21:
@gsalingu have you tested this docker image with a ssl-enabled postgres instance?

gsalingu:
Couldn't ever get the image buikt from GitHub 

I manually applied the patch and noticed it did not work

I'll go ahead and close the PR

Unfortunately the underlying JavaScript library does not seem to like ssl + postgres

PR #5016, opened 2025-12-05
Add pt-PT lang

maghuro:
Add Portuguese (European) language


jc21:
Can you please rebase and resolve conflicts? I would do it for you, but..
> These conflicts are too complex to resolve in the web editor.

PR #5114, opened 2025-12-31
Added logging for streams based on port

Shotz5:
I was looking at my service logs behind NPM and noticed some bad actors trying (and failing as I have it setup properly :D ) to access my service behind an NPM stream. I looked at the NPM logs and noticed that there was no place for stream-specific logs to grab the actors IPs and block them. So I've added it myself for contribution back. I'm sure I could have grabbed it from the NGINX logs within the container, but this is a much more NPM friendly way of logging imo.


Shotz5:
Been running on my local NPM server for over a week, working great, was able to trace IPs of malicious actors and add then to a block list. A whitelist filter for streams (similar to the one for proxy hosts) would be nice, perhaps I'll work on something for that in the future.

jc21:
Great stuff, thanks :)

PR #5118, opened 2026-01-03
Add DNS plugin for All-Inkl provider

mobilandi:
So far there was no certbot Plugin for the big German provider All-Inkl.
I [built that plugin](https://github.com/mobilandi/certbot-dns-kas) and tested it for Cert Requests via DNS-Challenge. Everything worked fine for new.
I thought it would be nice to add this to the main project. It's my first python project and also my first git PR. So feel free to criticize if something else needs to be done. Thank you.


mobilandi:
I'm confused. This was a runner error, not code error, right?

PR #5119, opened 2026-01-04
Added support for DNS challenges with Simply.com

manisto:
According to the documentation found at https://www.simply.com/en/docs/api/, a certbot plugin for their API is implemented at https://github.com/JohNan/certbot-dns-simply. This PR adds an entry to `dns-plugins.json` to enable the use of this plugin. Functionality has been tested with a locally built image, and works with let's encrypt.


grbrigsted:
Is there anything I can do to help speed up the verification process? I am already using the new provider with a locally built image, but it would be nice to be able to use it directly from Docker Hub.

jc21:
Since it's a new dns provider, and you've tested it, should be all good. Thanks

PR #5131, opened 2026-01-10
Feat: support fetching EdgeOne IP ranges

tianq02:
Lint issues fixed-ish this time
- Added EdgeOne IP ranges fetching support in backend/internal/ip_ranges_eo.js
- change real_ip_header from X-Real-IP to X-Forwarded-For in docker/rootfs/etc/nginx/nginx.conf
- add docs for the new feature


backend/index.js
```diff
@@ -3,26 +3,40 @@
...
+		    if (!EO_IP_RANGES_FETCH_ENABLED) {
+		        logger.info("EO IP Ranges fetch is disabled by environment variable");
+		    } else {
+		        logger.info("EO IP Ranges fetch is enabled");
+		        return internalIpRangesEO.fetch().catch((err) => {
+		            logger.error("EO IP Ranges fetch failed, continuing anyway:", err.message);
+		        });
+		    }
 		})
 		.then(() => {
```
tianq02: There's an severe bug, EO IP range never gets updated, and original IP ranges would run periodically even if disabled in env.

My fault copy and pasting existing code without double checking.

tianq02:
We found a bug, another pr would be sent after fixing, don't merge this.

PR #5177, opened 2026-01-16
Support configurable key sizes for Let’s Encrypt certificates

KalebCheng:
## Summary

This PR improves the Let’s Encrypt certificate workflow by adding support for **configurable key sizes** during certificate issuance and renewal.  
Users can now choose appropriate security levels based on their environment while keeping sensible defaults and full backward compatibility.

---

## Changes

- Added support for selecting **key type** (ECDSA / RSA) and corresponding **key sizes**:
  - ECDSA: 256 (default), 384
  - RSA: 2048 (default), 3072, 4096
- Key size options dynamically adapt to the selected key type, with smart defaults applied automatically.
- Key size configuration is consistently applied to both **certificate creation and renewal** (HTTP & DNS).
- Existing certificates remain unaffected, and unspecified key sizes fall back to Let’s Encrypt defaults.
- Full internationalization support across all supported languages.

---

## Testing

- Verified certificate issuance with multiple ECDSA and RSA key sizes.
- Verified key type switching correctly resets the default key size.
- Verified HTTP certificate renewal preserves the configured key parameters.

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -24,6 +24,18 @@ const certbotCommand = "certbot";
 const certbotLogsDir = "/data/logs";
 const certbotWorkDir = "/tmp/letsencrypt-lib";
 
+// Key type constants
+const KEY_TYPE = {
+	ECDSA: "ecdsa",
+	RSA: "rsa",
+};
+
+// ECDSA curve mapping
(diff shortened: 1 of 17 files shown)


CamelT0E:
Thank you for the very quick implementation.

PR #5182, opened 2026-01-17
Add sakuracloud dns provider

NAKNAO-nnct:
Added support for [Sakura Cloud](https://cloud.sakura.ad.jp/) DNS provider ( https://pypi.org/project/certbot-dns-sakuracloud/ )

### DIFF
--- a/backend/certbot/dns-plugins.json
+++ b/backend/certbot/dns-plugins.json
@@ -588,7 +588,7 @@
 		"package_name": "certbot-dns-sakuracloud",
 		"version": "~=5.2.2",
 		"dependencies": "",
-		"credentials": "# Sakura Cloud API credentials used by Certbot\ndns_sakuracloud_api_token = YOUR_API_TOKEN\ndns_sakuracloud_api_secret = YOUR_API_SECRET",
+		"credentials": "# Sakura Cloud API credentials\ndns_sakuracloud_api_token = YOUR_API_TOKEN\ndns_sakuracloud_api_secret = YOUR_API_SECRET",
 		"full_plugin_name": "dns-sakuracloud"
 	},
 	"timeweb": {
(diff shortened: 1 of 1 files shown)


O-hiron:
👍

Verified this PR using the build image `nginxproxymanager/nginx-proxy-manager-dev:pr-5182` against a real Sakura Cloud DNS zone. It works — wildcard certificate issued successfully through the UI.

- `*.<zone>` + `<zone>`, issued via DNS Provider "Sakura Cloud"
- TXT records were visible on the authoritative servers within ~30s, so the plugin's default 90s propagation is plenty — no tuning needed
- Cleanup removed the records correctly
- No dependency conflict: certbot core stays at the version shipped in the image
- The Sakura Cloud API key only needs the "modify settings" access level, not full create/delete

issued-certificate

dns-provider-dropdown

```
[Certbot] › ☒ complete  Installed sakuracloud
[SSL] › ℹ info  Requesting LetsEncrypt certificates via Sakura Cloud
Waiting 90 seconds for DNS changes to propagate
Successfully received certificate.
```

One suggestion on the entry itself: `certbot-dns-sakuracloud` is maintained by the Certbot project and released in lockstep with certbot core, so it can use the `{{certbot-version}}` substitution that `backend/lib/certbot.js` provides, like `route53` and the other first-party plugins do:

```json
"dependencies": "acme=={{certbot-version}}",
"version": "=={{certbot-version}}"
```

`~=5.2.2` matches the certbot in this PR build (5.2.2), but `develop` is on 5.6.0 now, so merging as-is would pin the plugin several minor versions behind and require manual bumping later.

The change is here if you'd like to take it directly:
https://github.com/O-hiron/nginx-proxy-manager/tree/fix-sakuracloud-version

If you'd rather not, or if you're no longer working on this, I'm happy to open a separate PR with the verification included — just let me know.

PR #5212, opened 2026-01-23
Feat: allow user set custom ports for Nginx and Web UI

beniamin:
Latest versions of TrueNAS have been migrated the Apps from k8s to docker and nginx-proxy-manager is the recommended way(in multiple discussion) to be used as a proxy for internal applications. For the setup to be secure and Nginx-proxy-manager to properly forward the traffic to the internal apps, it must be deployed using host_network options. 

There are several issues on TrueNAS Apps Catalog repo([[Enhancement]: Nginx Proxy Manager - host network](https://github.com/truenas/apps/issues/4149) and [Allow Host Networking for Nginx Proxy Manager](https://github.com/truenas/apps/issues/734)) asking for possibility to set nginx-proxy-manager with host_network, but since 80,443 and 81 ports may conflict with other running services on the server, TrueNAS admin require that nginx-proxy-manager should have the possibility to  to use custom ports instead of default ones.

This PR change current behavior and adds option for default ports to be changed trough HTTP_PORT, HTTPS_PORT and WEB_UI_PORT environment variables.


docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/60-ports.sh
```diff
@@ -0,0 +1,40 @@
...
+
+process_folder () {
+	echo "Processing folder: $1"
+	FILES=$(find "$1" -type f -name "*.conf")
+
+	for FILE in $FILES
+	do
+		echo "- ${FILE}"
+		# Replace HTTP port
+		sed -i -E "s/listen 80([^0-9]|$)/listen ${HTTP_PORT}\1/g" "$FILE"
```
jc21, line 21: I see what you're trying to do here and I'm glad you thought of this. Something I would consider though..

This script runs on each startup, yet it seems like a run-once migration. Some options that might be better:
1. create a migration, they're not just for databases. In this case it could iterate over all hosts and regenerate their nginx config, however there are risks doing this too
2. Or keep this file but don't be specific with 80/443. Consider what should happen if I change to 8080/8443 and then change to 8888/9999. I'd expect this script to handle it

Option 2 is probably my preference here.

jc21:
Thanks for the work on this! It's been on my mind to do this for a while...

Delivator:
Any updates on this?

beniamin:
@jc21 I had some time this weekend to update the PR. Now it should work with any port that was previously set. 
Also, I tried to add a new test for this use case. Hope this works

beniamin:
Is there anything I need to change on this?

PR #5268, opened 2026-02-03
Clarify network_mode: host ports

soundstorm:
Just deployed npm on a machine, where an orphaned grafana server was serving on port 3000. When switching to `network_mode: host` the npm API-service would silently fail and the frontend loaded the grafana frontend instead. To help people running other services possibly occupying port 3000 and therefor the npm API, I added the short line in the quick setup.
Might be a good addition to the docs as well.


DominicWrege:
I have the same problem. I need the 3000 port for something else.

PR #5282, opened 2026-02-06
Add Added the option to use different access lists for custom locations

MioOgbeni:
## Summary

This PR adds support for custom access lists (ACLs) per proxy host location, allowing users to override the parent proxy host's ACL for specific locations.

## Problem Context

In nginx-proxy-manager v2.3.1 (very old version that we are using now), users could achieve similar functionality through advanced config by pasting ACL rules into location advanced config. However, when we decide to upgrade to newer versions, we noticed that this approach became problematic because:

- The new version always adds the parent proxy host's ACL with a `deny all` rule at the end
- This `deny all` rule is rendered before any advanced config at location block, making all other ACL rules ineffective
- Users lost the ability to have different access controls for different locations within the same proxy host

## Solution

Instead of relying on advanced config workarounds, this PR implements a proper UI-driven approach that allows users to:
- Enable/disable inheritance of the parent proxy host's ACL for each custom location
- Select a different ACL specifically for that location when inheritance is disabled

### Backend Changes

**Schema Updates:**
- Added `location_count` to access list schema for separate counting of access list usage across custom locations of proxy hosts
- Updated proxy host location schema to include `use_parent_access_list`, `access_list_id`, and `access_list` properties

**Core Logic (`backend/internal/`):**
- `access-list.js`: Split ACL counting into `proxy_host_count` and `location_count` for better visibility/auditability
- `nginx.js`: Fixed ACL rendering logic to properly apply location-specific ACLs with correct precedence
- `proxy-host.js`: Added location normalization and ACL enrichment functions

### Frontend Changes

**UI Components:**
- `LocationsFields.tsx`: Added checkbox to toggle parent ACL inheritance and AccessField dropdown for custom ACL selection
- `Table.tsx`: Added separate "Locations" column showing location-specific ACL usage counts

**API Models:**
- `models.ts`: Added `locationCount` property to AccessList interface

**Translations:**
- Added "locations" and "locations.count" keys to all 18 supported languages

### API Schema Updates
- Updated all access list endpoint schemas to include the new `location_count` property

## Testing

- All existing functionality remains intact
- New feature works correctly with proper ACL inheritance and overrides
- ACL usage counts are accurately split between proxy hosts and locations
- ACL edits/deletes properly re-render affected locations
- Schema validation passes for all endpoints

## Notes

This is my first contribution to the nginx-proxy-manager project. I'm very open to any criticism and code reviews - please let me know if there are any improvements, bugs, or better approaches to implement this feature. The code has been tested in the development environment and appears to work correctly (at least i hope 🤞), but I'm happy to make any necessary adjustments.

### DIFF
--- a/backend/internal/access-list.js
+++ b/backend/internal/access-list.js
@@ -15,6 +15,47 @@ const omissions = () => {
 	return ["is_deleted"];
 };
 
+// Build a map of ACL id -> usage count from locations only (host-level usage is counted by SQL)
+const buildAccessListUsageMap = async () => {
+	const usage = {};
+	const hosts = await proxyHostModel.query().select("locations").where("is_deleted", 0);
+	hosts.forEach((host) => {
+		if (Array.isArray(host.locations)) {
+			host.locations.forEach((loc) => {
+				if (loc && loc.use_parent_access_list === false && loc.access_list_id) {
+					usage[loc.access_list_id] = (usage[loc.access_list_id] || 0) + 1;
+				}
+			});
+		}
+	});
+	return usage;
+};
+
+// Find proxy hosts that reference an ACL either directly or via locations
+const findProxyHostsUsingAcl = async (accessListId) => {
+	const hosts = await proxyHostModel.query().where("is_deleted", 0);
+	return hosts.filter((host) => {
+		if (host.access_list_id === accessListId) return true;
+		if (Array.isArray(host.locations)) {
+			return host.locations.some(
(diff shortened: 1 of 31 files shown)


MioOgbeni:
Hi @jc21, is there any way you could take a look at this PR?

It's a new feature, so I'm not sure if it will be possible to merge it. But it doesn't drastically change the current functionality, it just adds the option to have different ACLs for individual proxy hosts and their custom locations.

PR #5275, opened 2026-02-04
Add reuseport to improve performance and reduce packet drops

7heMech:
Users reported extreme packet loss (up to 80%) when streaming UDP-based services like TeamSpeak 3 through Nginx Proxy Manager. This issue is primarily caused by lock contention on the UDP receive buffer when multiple Nginx worker processes compete for the same socket.

This PR enables the reuseport socket option for both TCP and UDP stream templates.

By enabling reuseport, the Linux kernel creates multiple listening sockets and balances incoming packets across workers at the kernel level. This ensures that packets from the same connection consistently reach the same worker, significantly reducing packet drops and stabilizing high-bandwidth UDP streams.

The change is compatible with Linux Kernel 3.9+ (released in 2013) or Debian 7.

Resolves #5000

Thanks to @zdzichu6969

Issue #2834, opened 2023-04-15
Support HTTP/3


joaopinto14:
Add HTTP protocol version 3 support with quic

parnavh:
See similar thread: #1550

WhippingBoy01:
Vote from me

barnhill:
This would be a great addition

vincent1890:
up

CyberCreator:
I support it. It's an interesting idea.

barnhill:
Dont let this go stale bot!

Timmmel:
Voting

MattXcz:
Voting

kylhuk:
Technically it should support HTTP/3 already. OpenResty added support in v1.27.1.1 and nginx-proxy-manager uses v1.27.1.2. So it should be only a configuration issue.

On this line: https://github.com/NginxProxyManager/nginx-proxy-manager/blob/534afe606708ee5ffe09bd67ccdbf91b58a7e956/docker/rootfs/etc/nginx/conf.d/default.conf#L25 you would have to add `http2` and `quic` to enable full support. Something like:
```
...
	listen 443 ssl http2;
	listen [::]:443 ssl http2;
	listen 443 quic reuseport;
	listen [::]:443 quic reuseport;

	ssl_protocols TLSv1.3;
...
```
If I find time I will open up a PR

Also, duplicate of #1550

danfratamico:
> Technically it should support HTTP/3 already. OpenResty added support in v1.27.1.1 and nginx-proxy-manager uses v1.27.1.2. So it should be only a configuration issue.
> 
> On this line:
> 
> [nginx-proxy-manager/docker/rootfs/etc/nginx/conf.d/default.conf](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/534afe606708ee5ffe09bd67ccdbf91b58a7e956/docker/rootfs/etc/nginx/conf.d/default.conf#L25)
> 
> Line 25 in [534afe6](/NginxProxyManager/nginx-proxy-manager/commit/534afe606708ee5ffe09bd67ccdbf91b58a7e956)
> 
>  listen 443 ssl; 
> you would have to add `http2` and `quic` to enable full support. Something like:
> ```
> ...
> 	listen 443 ssl http2;
> 	listen [::]:443 ssl http2;
> 	listen 443 quic reuseport;
> 	listen [::]:443 quic reuseport;
> 
[...]

Issue #1550, opened 2021-11-02
HTTP3 + QUIC support


nyanmark:
With QUIC support being in most of today's browsers, and tools already excising to enable QUIC within nginx I believe it would be a good idea to add the support to NPM helping users speed up those huge video streams.

iamgadgetman:
It would be nice to be able to use NPM to connect to internet radio servers that use QUIC.

ingarb:
Hope this can be added to NPM. I would imagine if you don't have QUIC Google might rank your site lower copared to a site that does have QUIC. It would also make your site a little fit faster - also leading to a higher ranking.

danfratamico:
I totally agree here, this needs to be a top priority.

timnolte:
I was going about the route of using NPM in front of a Docker/container setup only to realize that although the web server I'm running should support QUIC & HTTP/3 unfortunately NPM doesn't. Having NPM in front of any containers/services I start up is very handy but I'd sure like the ability to use QUIC & HTTP/3. Keeping an eye on this.

nonlinearsugar:
It's still experimental/beta in Nginx and only available as separate packages/in a separate branch. https://www.nginx.com/blog/binary-packages-for-preview-nginx-quic-http3-implementation/ https://github.com/nginx-quic/nginx-quic

While I lend my support to those asking, I don't think this can/should go anywhere on the NPM side before Nginx merge it into their stable branch.

EnfermeraSexy:
> It's still experimental/beta in Nginx and only available as separate packages/in a separate branch. https://www.nginx.com/blog/binary-packages-for-preview-nginx-quic-http3-implementation/ https://github.com/nginx-quic/nginx-quic
> 
> While I lend my support to those asking, I don't think this can/should go anywhere on the NPM side before Nginx merge it into their stable branch.

It's already merged, right?

nonlinearsugar:
Yeah.  Time permitting, I think this is ready for @jc21 to action.

https://nginx.org/en/docs/quic.html

squadramunter:
Any update on this?

multipliedtwice:
Would be amazing to have it

stanthewizzard:
would be awesome ... no news ?

LZNetwork-Main:
I hope we will see this added in the near future, I am already using NGINX with HTTP/3 on some of my servers and it can improve things significantly with larger sites!
[...]

PR #5296, opened 2026-02-09
Support for HTTP/3

kylhuk:
Added support for HTTP/3 since the OpenResty version supports it, so it was only a matter of adding it to the Nginx config templates and adding it to the frontend.

I am still testing, which is why this is a draft.


kylhuk:
Affects the following issues:
#2834
#1550

toviszsolt:
@kylhuk 
- I would like to draw your attention to the fact that PR is still in draft mode.
- Can you please confirm that you have tested the build?
`nginxproxymanager/nginx-proxy-manager-dev:pr-5296`

kylhuk:
@toviszsolt Have not tested it yet, which is why it is still in draft mode. Will update you as soon as the PR is ready

toviszsolt:
@kylhuk 
- If you feel that your PR is ready and would like help with testing, I'm happy to help, just let me know.
- Could you perhaps ask CoPilot for a review?

Steve-Tech:
Hi, I tested this, I seem to be getting: `{"nginx_online":false,"nginx_err":"nginx: [emerg] unknown directive \"http3\" in /data/nginx/proxy_host/1.conf:27\nnginx: configuration file /etc/nginx/nginx.conf test failed\n"}`

I don't know if this error is printed anywhere reasonable, but it is saved in the database: `select meta from proxy_host;`.

I think you will have to add `--with-http_v3_module` here: https://github.com/NginxProxyManager/docker-nginx-full/blob/master/scripts/build-openresty

jc21:
> I think you will have to add --with-http_v3_module here: https://github.com/NginxProxyManager/docker-nginx-full/blob/master/scripts/build-openresty

I've done this on the `http3` branch:
https://github.com/NginxProxyManager/docker-nginx-full/commit/f7a265b2f7f68578bdf2ad5a4d4e30d8e73b44de

Just waiting for a successful build before merging to master.

toviszsolt:
@jc21
Sounds good, I'm really looking forward to http3 being turned on. Let's do it, guys, go for it! If you need help, I'm here for you!

jc21:
Just testing this image, a few things going on.

1. When enabling http3 in the ui, the app reports it's successfully saved but something is going wrong, for me, and the conf file is removed from disk. Disabling the option and saving is actually successful though
2. The UI for the buttons is weird now and needs some love
3. The nginx template should also add the following to the server block when http3 is enabled:
```
# Advertise HTTP/3 support to clients
add_header Alt-Svc 'h3=":443"; ma=86400';
```

jc21:
Ok so there's a big problem as is.

```
[2/17/2026] [4:34:04 AM] [Nginx    ] › ⬤  debug     Nginx test failed: nginx: [emerg] duplicate listen options for 0.0.0.0:443 in /data/nginx/proxy_host/49.conf:23
nginx: configuration file /etc/nginx/nginx.conf test failed
[2/17/2026] [4:34:04 AM] [Nginx    ] › ⬤  debug     Deleting file: /data/nginx/proxy_host/3.conf
```

To replicate, enable http3 on any host. Then do it on another host.

```
listen 443 quic reuseport;
```

This doesn't like to be defined more than once across multiple server blocks. AI tells me:

>reuseport only makes sense when multiple sockets are bound to the same 4‑tuple (IP:port + protocol family) so the kernel can distribute incoming packets across those sockets. With HTTP/3 (QUIC over UDP) in nginx, one UDP socket per address:port is effectively “owned” by a single ⁠server context (and nginx’s QUIC implementation expects to manage QUIC connection state off that one socket). If you try to declare ⁠listen 443 quic reuseport; in multiple ⁠server blocks for the same address:port, nginx treats it as an invalid/ambiguous configuration (multiple competing UDP QUIC listeners on the same endpoint), so it won’t allow it.

> What to do instead
	•	Define the QUIC listener once (in a single ⁠server block) and route by SNI/Host as usual via ⁠server_name. 	•	Keep your TCP/TLS HTTP/1.1+HTTP/2 ⁠listen 443 ssl; on multiple server blocks if you want—the QUIC ⁠listen ... quic is the one that must not be duplicated for the same IP:port.

If I'm being honest, this doesn't seem feasible given the current ecosystem of nginx config generation.

Steve-Tech:
Could we put `listen 443 quic reuseport;` in [docker/rootfs/etc/nginx/conf.d/default.conf](https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/docker/rootfs/etc/nginx/conf.d/default.conf) and then use `listen 443 quic;` for the proxy hosts?

E.g: https://stackoverflow.com/a/77005737

kylhuk:
Addressed the `reuseport` issue and added it to `default.con` and removed it from templates (38641bb9542a94e69b4c2edcd9dae8b1b47fc3b7)

kylhuk:
@jc21 Feel free to modify this PR and or move it to a branch on your repo. Depending how big the `reuseport` issue is of course :)

docker/rootfs/etc/nginx/conf.d/default.conf
```diff
@@ -22,8 +22,8 @@ server {
 
 # First 443 Host, which is the default if another default doesn't exist
 server {
-	listen 443 ssl;
-	listen [::]:443 ssl;
+	listen 443 ssl reuseport;
+	listen [::]:443 ssl reuseport;
```
Steve-Tech, line 26: I think you should add a new line for each of the `reuseport` entries, I'm not sure what the implications and behaviour are with tacking it on the end of `listen 443 ssl` instead of `listen 443 quic`.

Unless you're sure this is fine.

kylhuk: I am relying on the documentation of nginx and this says:

```
Syntax: 	listen address[:port] [default_server] [ssl] [http2 | quic] [proxy_protocol] [setfib=number] [fastopen=number] [backlog=number] [rcvbuf=size] [sndbuf=size] [accept_filter=filter] [deferred] [bind] [ipv6only=on|off] [reuseport] [so_keepalive=on|off|[keepidle]:[keepintvl]:[keepcnt]];
            listen port [default_server] [ssl] [http2 | quic] [proxy_protocol] [setfib=number] [fastopen=number] [backlog=number] [rcvbuf=size] [sndbuf=size] [accept_filter=filter] [deferred] [bind] [ipv6only=on|off] [reuseport] [so_keepalive=on|off|[keepidle]:[keepintvl]:[keepcnt]];
            listen unix:path [default_server] [ssl] [http2 | quic] [proxy_protocol] [backlog=number] [rcvbuf=size] [sndbuf=size] [accept_filter=filter] [deferred] [bind] [so_keepalive=on|off|[keepidle]:[keepintvl]:[keepcnt]];
Default: 	

listen *:80 | *:8000;

Context: 	server
```
Source: https://nginx.org/en/docs/http/ngx_http_core_module.html#listen

According to that, `listen 443 ssl reuseport;` should be correct. But I have not tested it

toviszsolt:
I have also done some research.

**Snippets would be needed for http3:**
`snippets/http3-first.conf`
```
# Use this ONLY in the first server block on :443/udp
listen 443 quic reuseport;
listen 443 ssl;
add_header Alt-Svc 'h3=":443"; ma=86400' always;
```

`snippets/http3.conf`
```
# Use this in all other server blocks on :443/udp
listen 443 quic;
listen 443 ssl;
add_header Alt-Svc 'h3=":443"; ma=86400' always;
```
> Reuseport should only be specified once for the same IP:port combination. If you use it in the first server block with `listen 443 quic reuseport;`, all other server blocks on the same port should use `listen 443 quic;` without the reuseport directive. The operating system-level socket option will be shared across all worker processes and virtual hosts automatically, avoiding potential conflicts.

**And it should be placed in the `server` block when http3 is enabled:**
```
# First server block
server {
    include snippets/http3-first.conf;
    
    server_name example.tld;
    # ...
}

# All other server blocks
server {
    include snippets/http3.conf;
    
    server_name other.tld;
    # ...
}
```

**It is important that Stream Hosts do not allow the option to enable http3**
- For HTTP based hosts (Proxy, Redirection, 404), http3 should work, as an HTTP response is sent.
- For Stream, TCP/UDP forwarding happens instead of sending an HTTP response.

**Additional fine-tuning parameters for the http3 protocol:**
These parameters should be placed in the `http` block if you wish to set them.
```
http {
    # [RECOMMENDED]
    # Enable QUIC retry packets to validate client addresses 
    # and prevent IP spoofing attacks
    quic_retry on;
    
    # [SECURITY RISK]
    # Allow 0-RTT (zero round-trip time) early data for faster connection resumption
    # Warning: This may be vulnerable to replay attacks, use with caution
    ssl_early_data on;
    
    # [RECOMMENDED]
    # Enable Generic Segmentation Offloading for QUIC to improve performance
    # by offloading packet segmentation to the network card
    quic_gso on;
    
    # [RECOMMENDED]
    # Set buffer size for HTTP/3 request and response streams
    # Larger values may improve performance for big uploads/downloads
    http3_stream_buffer_size 64k;
}
```

docker/rootfs/etc/nginx/conf.d/default.conf
toviszsolt: Hi guys, I've written a more detailed comment here, I hope it will help: https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5296#issuecomment-3915096086

Steve-Tech: Yes it's syntactically correct, however SSL is TCP, while QUIC is UDP. So I don't think it'll be passed to the correct bind call.

kylhuk: > Yes it's syntactically correct, however SSL is TCP, while QUIC is UDP. So I don't think it'll be passed to the correct bind call.

Ah! Now I see what you are saying, and you are probably correct. The answer of @toviszsolt also supports your argument.

kylhuk:
Thanks @toviszsolt this was very helpful. Tried to implement a logic, so that we have a different config for the first http server. I did not test the new image yet, so I am not sure if my change affects also the Stream servers. The SSL buttons get rendered in `frontend/src/components/Form/SSLOptionsFields.tsx`

backend/internal/certificate.js
```diff
@@ -630,7 +630,7 @@ const internalCertificate = {
 	 * @param {String}  privateKey    This is the entire key contents as a string
 	 */
 	checkPrivateKey: async (privateKey) => {
-		const filepath = await tempWrite(privateKey, "/tmp");
+		const filepath = await tempWrite(privateKey);
```
toviszsolt, line 633: This change seems like a fair bug fix. The second parameter is optional, although lib does not specify where the temp file will be created. Reference: https://www.npmjs.com/package/temp-write

backend/internal/certificate.js
```diff
@@ -660,7 +660,7 @@ const internalCertificate = {
 	 * @param {Boolean} [throwExpired]  Throw when the certificate is out of date
 	 */
 	getCertificateInfo: async (certificate, throwExpired) => {
-		const filepath = await tempWrite(certificate, "/tmp");
+		const filepath = await tempWrite(certificate);
```
toviszsolt: See my previous comment on line 633

backend/templates/_listen.conf
toviszsolt, line 1: The logic seems correct.

toviszsolt (approved):
I have reviewed all the changes and everything looks correct. I have left some comments on the modified files, which are just for information to help with the review process.

7heMech:
> Ok so there's a big problem as is.
...
> > What to do instead
> > •	Define the QUIC listener once (in a single ⁠server block) and route by SNI/Host as usual via ⁠server_name. > 	•	Keep your TCP/TLS HTTP/1.1+HTTP/2 ⁠listen 443 ssl; on multiple server blocks if you want—the QUIC ⁠listen ... quic is the one that must not be duplicated for the same IP:port.



Can't we do something like that? Here's what AI told me :joy: 

---


The issue is that Nginx treats `reuseport` as a socket **creation** instruction, not just a property. You cannot tell Nginx to "create" the same socket (0.0.0.0:443 UDP) multiple times with that flag.

Here is the exact fix you need to apply to the code to get this working, along with the concern regarding the "Stream" PR.

### Part 1: The Fix for HTTP/3 (PR #5296)

You need to split the configuration into two parts: one "Initial" file that loads first and claims the `reuseport` capability, and then the Template which just attaches to that existing port.

#### Step 1: Create an Initialization File

You need a file that loads **before** your proxy hosts (which usually start with numbers like `1.conf`, `2.conf`). Nginx loads files alphabetically.

Create a file named `/etc/nginx/conf.d/00-http3-init.conf` (or inside your docker mapping) with this content:

```nginx
# 00-http3-init.conf
# This block exists ONLY to initialize the QUIC socket with reuseport.
# It handles requests for unknown domains or IP-direct access.

server {
    listen 443 quic reuseport; 
    listen [::]:443 quic reuseport; 
    
    # SSL Config is required for QUIC to start
    # Point these to your fallback cert or any valid dummy cert
    ssl_certificate /etc/nginx/certs/dummy.crt; 
    ssl_certificate_key /etc/nginx/certs/dummy.key;
    
    server_name _;
    
    # Return 444 (No Response) to drop connections that hit this default block
    return 444; 
}

```

#### Step 2: Modify the Template

Now, modify the file `backend/templates/_listen.conf`. This is the snippet Nginx Proxy Manager uses to generate every host's config.

**Current (Broken):**

```nginx
listen 443 quic reuseport;
listen [::]:443 quic reuseport;

```

**New (Fixed):**

```nginx
# Remove 'reuseport' here because 00-http3-init.conf already opened the socket
listen 443 quic;
listen [::]:443 quic;

```

By removing `reuseport` from the template, every new host you add will simply "attach" to the existing QUIC listener created by step 1, preventing the duplicate error.

---

### Part 2: Is the other PR (#5275 - Streams) safe?

You asked if **PR #5275 (Add reuseport to Streams)** has the same risk.

**Verdict: It is safe, but requires caution.**

The reason HTTP/3 crashes and Streams (usually) won't is due to **Port Uniqueness**.

* **HTTP/3 Issue:** Every single website you host listens on **Port 443**. Since 50 websites all try to claim "reuseport on 443", they collide.
* **Streams Issue:** In Nginx Proxy Manager, "Streams" are typically used to forward specific ports (e.g., Minecraft on 25565, SQL on 3306).
* Stream A listens on Port 8000.
* Stream B listens on Port 8001.



Because the **ports are different**, `reuseport` is allowed on both. Nginx sees them as completely separate sockets.

**The Danger Zone for Streams:**
If you create **two** Stream hosts that listen on the **same port** (e.g., two UDP streams listening on port 5000, trying to rely on SNI/preread to route traffic), **PR #5275 will crash** with the exact same error.

**Summary Recommendation:**

* **For HTTP/3 (PR #5296):** You **must** implement the "Init File + Clean Template" fix above.
* **For Streams (PR #5275):** It is safe to merge **if** you guarantee that you never assign the same incoming port to multiple streams. If you do, it will crash Nginx.

7heMech:
Maybe we can even enable it globally like this, which will fix the streams issue.

```nginx

# 00-init.conf
# This initializes BOTH TCP and UDP sockets with reuseport.

server {
    # 1. HTTP/1.1 (TCP)
    listen 80 reuseport;
    listen [::]:80 reuseport;
    
    # 2. HTTPS/2 (TCP)
    listen 443 ssl http2 reuseport;
    listen [::]:443 ssl http2 reuseport;
    
    # 3. HTTP/3 (UDP) - This is the missing part!
    listen 443 quic reuseport;
    listen [::]:443 quic reuseport;
    
    # SSL Config (Required for both TCP/443 and UDP/443)
    # Point these to any valid cert/key pair you have
    ssl_certificate /etc/nginx/certs/dummy.crt;
    ssl_certificate_key /etc/nginx/certs/dummy.key;
    
    # Advertise HTTP/3 availability to browsers (Optional)
    add_header Alt-Svc 'h3=":443"; ma=86400';
    
    server_name _; # Matches any request not caught by other blocks
    return 444;    # Drop the connection
}

```

toviszsolt:
@7heMech 
I'm sorry, but it seems you haven't read the history. The issue you mentioned has already been discussed and resolved. Can you please read all the previous comments on PR? Thanks.

Update: Please note that this PR does not affect Stream Hosts!

7heMech:
> @7heMech I'm sorry, but it seems you haven't read the history. The issue you mentioned has already been discussed and resolved. Can you please read all the previous comments on PR? Thanks.
> 
> Update: Please note that this PR does not affect Stream Hosts!

Hello, friend, I'm saying this because I have a related PR which has to implement reuseports as a performance bug fix for streams, and here we have the opportunity to improve performance for everything, it would be beneficial for HTTP3 as well since it's over UDP it will likely have similar issues as with streams losing 80% of data.

toviszsolt:
@7heMech 
Thanks, but your PR doesn't affect this PR and vice versa. We have already discussed this in previous comments. Your PR only affects `Stream Hosts`.

_Related part of `backend/templates/_listen.conf`_
```
{% if tcp_forwarding == 1 or tcp_forwarding == true -%}
{% if udp_forwarding == 1 or udp_forwarding == true -%}
```

This PR omits `Stream Hosts`, precisely and it is intentionally NOT possible to enable http3 support for `Stream Hosts` in the UI. HTTP3 can only be enabled in the UI in the case of `Proxy`, `Redirection`, and `Dead Hosts`.

_Related part of `backend/templates/_listen.conf`_
```
{% if http3_support == 1 or http3_support == true %}
```

Modified files related to frontend modals:
- `frontend/src/modals/DeadHostModal.tsx`
- `frontend/src/modals/ProxyHostModal.tsx`
- `frontend/src/modals/RedirectionHostModal.tsx`

**NOT** modified files related to frontend modals:
- `frontend/src/modals/StreamModal.tsx`

7heMech:
> @7heMech 
> Thanks, but your PR doesn't affect this PR and vice versa. We have already discussed this in previous comments. Your PR only affects `Stream Hosts`.

I didn't say they have something in common... I'm saying we should keep reuseport here like you originally intended just in another manner of activation.

toviszsolt:
@7heMech
I don't think the global settings you mentioned are the responsibility of this PR.
This PR deals with the implementation of enabling the HTTP/3 module and is not responsible for global settings outside the HTTP/3 protocol or Stream Host issues. Moreover, this PR does not even affect Stream Hosts. I understand that you are trying to solve a issue related to Stream Hosts, but this PR does not address a bug, but rather a feature.

7heMech:
> @7heMech
> I don't think the global settings you mentioned are the responsibility of this PR.
> This PR deals with the implementation of enabling the HTTP/3 module and is not responsible for global settings outside the HTTP/3 protocol or Stream Host issues. Moreover, this PR does not even affect Stream Hosts. I understand that you are trying to solve a issue related to Stream Hosts, but this PR does not address a bug, but rather a feature. 
> 

Am I talking to AI 🤣

It's clearly related since it was discussed here. But I'll take it elsewhere maybe, @jc21 can I have feedback on the global snippet I suggested?

Edit: My suggestion here is only for http3, but I'm saying maybe it'll be good to implement globally in separate PR.

Steve-Tech:
@7heMech, Like @toviszsolt said this doesn't effect Streams as that uses dedicated ports, whereas http3/quic is all port 443, and the `quic reuseport` issue has already been resolved (but I am yet to test). Also your comment with `00-init.conf` isn't too helpful since we already know what we need to do, and it would conflict with the "You are not configured" page or whatever default page is enabled, adding logic to `_listen.conf` does seem like the best option and is what's already done.

Edit: Although I'm not sure what would happen if the `http3_first` host were deleted, does it just regenerate all the configs and pick a new `http3_first`?

7heMech:
> adding logic to `_listen.conf` does seem like the best option and is what's already done.

Right, somehow I missed that...

diogosilvan:
Is there any ETA for HTTP/3 support to be available via the control panel in the stable version of jc21/nginx-proxy-manager:latest?

kylhuk:
> Edit: Although I'm not sure what would happen if the `http3_first` host were deleted, does it just regenerate all the configs and pick a new `http3_first`?

Yes, the http3_first is the first to be generated. So if I delete this one, another one will become http3_first automatically. There should technically always be a http3_first.

jc21:
> Is there any ETA for HTTP/3 support to be available via the control panel in the stable version of jc21/nginx-proxy-manager:latest?

No. There's never an ETA for pr's like this. This is a delicate change and TBH I'm not sold on the http3_first idea at all. As someone mentioned, what happens when that host is deleted?

I feel like I really need a solid 40 hour week to review this change and consider all the angles. Just because things work for you doesn't mean they work for all.

Also consider there's another PR that allows users to change the http/https port in the container, and it's a change I'd like to get in too.

htkoca:
I wonder if these cases we're talking about could be converted into tests, since this is a big feature we're trying to adopt

voltuer:
hi, was this ever merged to the main project?

7heMech:
> hi, was this ever merged to the main project?

As you can see the PR is open

PR #5302, opened 2026-02-13
I noticed the footer year was hardcoded to 2025. I've updated it to use a dynamic year so it stays current automatically

codetheuri:


codetheuri:
👍

PR #5316, opened 2026-02-16
Add SLSA generic generator workflow

rizuashafiqi93:
This workflow generates SLSA provenance files for projects, satisfying level 3 requirements. It includes steps for building artifacts and generating provenance subjects.


jc21:
I don't entirely see the point of this workflow. This project doesn't publish artifacts, only a docker hub image that I build in my own CI, not in a github action.

PR #5317, opened 2026-02-17
Fix uploading of custom certificates

### LINKED ISSUE #5086: unable to upload custom certificates
I want to upload custom certifcates but al i get is a message 'filepath is not defined' in the logs. The Ui gives 'Key files protected with a passphrase are not supported.' which is not true, also an text 'internal error' appears..
if I look in /tmp i see a lot of 00499d91-c59a-4f20-b067-5b4326aa6ced like dirs with the uploaded certs in it

Using the default docker-compose.yaml

**Checklist**
  - Yes
  - Yes 
  - Yes 

**Describe the bug**

**To Reproduce**
1. Go to certificate
2. add custom certificate
3. fill in the files
4. click saver

the certificate should be added

Docker-ce

Tech-no-1:
Fixes https://github.com/NginxProxyManager/nginx-proxy-manager/issues/5086

You do not have to specify a file path. In this case tmpWrite creates a temporary file with a random name and returns the path to that file. The file path is appended to the random path used by tmpWrite, so it must be relative. Something must have changed as it apparently worked before.

I created a test build with these changes and was able to successfully add a custom certificate.

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -630,7 +630,7 @@ const internalCertificate = {
 	 * @param {String}  privateKey    This is the entire key contents as a string
 	 */
 	checkPrivateKey: async (privateKey) => {
-		const filepath = await tempWrite(privateKey, "/tmp");
+		const filepath = await tempWrite(privateKey);
 		const failTimeout = setTimeout(() => {
 			throw new error.ValidationError(
 				"Result Validation Error: Validation timed out. This could be due to the key being passphrase-protected.",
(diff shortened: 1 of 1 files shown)


jc21:
Interesting. I literally just uploaded custom certs myself using the latest image, without any issues.

I don't mind merging this though.

sun-god-gear:
Just so you know, v2.15.1 still has this issue. I was able to workaround by inserting directly into the sqlite

PR #5324, opened 2026-02-18
chore: added Norwegian translation

biodland:
Added an Norwegian translation of the locale files.


biodland:
image

Seems to be working here, and showing in the menu and ui has now norwegian language.

PR #5331, opened 2026-02-19
Update EuroDNS plugin version to 1.8.0

creekorful:


jc21:
Anyone feel like testing the image?

DaanSelen:
> Anyone feel like testing the image?

I can if there is still someone listening?

Also: we can bump it to: https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5407

creekorful:
I have the image running in production for 2months but with an hardcoded fix in the dns-plugin.json.

I can confirm this is working.

Regards

DaanSelen:
> I have the image running in production for 2months but with an hardcoded fix in the dns-plugin.json.
> 
> I can confirm this is working.
> 
> Regards

I have done the same and it works nicely.

I only have to change the base-url. But thats ok for now. Hopefully @jc21 can take a look.

creekorful:
Indeed! I did remove completely the endpoint URL from the configuration.

Maybe it's worth removing or updating as well. Or we can drop my PR in favor of the one from Vincent. (#5407)

DaanSelen:
> Indeed! I did remove completely the endpoint URL from the configuration.
> 
> Maybe it's worth removing or updating as well. Or we can drop my PR in favor of the one from Vincent. (#5407)

Maybe, yeah, but I do not want to lose the attention of our beloved dev.

jc21:
Conflicts with #5535 which is more up to date

PR #5348, opened 2026-02-23
feat: fix upstream host resolution edge case and support PEM paste in custom certificates

### LINKED ISSUE #5344: proxy host upstream `host.docker.internal` cannot be resolved when `extra_hosts` is used (variable `proxy_pass` + resolver path)
## Summary

When Nginx Proxy Manager is deployed with Docker Compose and `extra_hosts` defines `host.docker.internal`, proxy host validation can still fail with:

`host.docker.internal could not be resolved (3: Host not found)`

This happens even though the hostname is present in the container's `/etc/hosts`.

## Version

- Reproduced on: `v2.13.7`
- The upstream resolution path appears unchanged in newer templates.

## Environment

- Docker Compose deployment
- `extra_hosts` configured for NPM container, e.g.:
  - `host.docker.internal:host-gateway` (or an explicit host IP)

## Steps to Reproduce

1. Deploy NPM with `extra_hosts` containing `host.docker.internal`.
2. Create a Proxy Host with:
   - Forward Hostname/IP: `host.docker.internal`
   - Forward Port: `<your service port>`
3. Save/apply the host config.

## Actual Behavior

NPM/Nginx config test fails with:

`host.docker.internal could not be resolved (3: Host not found)`

## Expected Behavior

If `host.docker.internal` is available via system resolver path (`/etc/hosts`), the proxy host should be accepted and work.

## Why this happens

NPM proxy templates use variable-based `proxy_pass` (e.g. `$server`), which relies on Nginx runtime DNS resolver behavior. That resolver does not use `/etc/hosts`, so `extra_hosts` entries are not honored in this path.

## Proposed Fix (backward-compatible)

Add an optional NPM-side pre-resolution step (disabled by default), for example:

- New env flag: `NPM_PRE_RESOLVE_UPSTREAM_HOSTS=true`
- During config generation:
  - pre-resolve eligible upstream hostnames via system resolver (`dns.lookup`)
  - write resolved IP into generated upstream target
  - if resolution fails, keep original hostname

This preserves current default behavior while providing a compatible path for Docker setups relying on `extra_hosts`.

## Notes

I can open a PR with implementation and docs updates. Default behavior remains unchanged unless the new env flag is enabled.

### LINKED ISSUE #5347: Feature request: allow pasting certificate/key PEM content directly in the Custom Certificate modal
## Feature request

Please add support for directly pasting PEM content in the **Custom Certificate** modal, instead of requiring file selection from the local machine.

## Current behavior

When creating a custom certificate, users are expected to upload files for:

- Certificate
- Certificate Key
- (Optional) Intermediate Certificate

In some environments (remote browser sessions, mobile devices, thin clients), choosing local files is inconvenient or not possible.

## Requested behavior

In addition to existing file upload inputs, provide text areas to paste:

- Certificate PEM
- Private Key PEM
- Intermediate Certificate PEM (optional)

Either input method should work:

- file upload, or
- pasted text

## Why this helps

- Better usability in remote/admin workflows
- Works on devices without easy file access
- Faster copy/paste from secret managers or vault tools

## Backward compatibility

Keep existing file-upload behavior unchanged; this should be additive only.

Thanks!

Eric-Terminal:
This PR addresses two issues:

- Closes #5344
- Closes #5347

### 1) Optional upstream pre-resolution for `forward_host` (issue #5344)

Problem:
Nginx variable-based upstream resolution may fail for hostnames that are only resolvable through container/system resolver paths (for example `extra_hosts` mappings like `host.docker.internal`).

What changed:
- Added optional env flag: `NPM_PRE_RESOLVE_UPSTREAM_HOSTS` (default: disabled)
- During Nginx config generation, NPM can pre-resolve eligible upstream hostnames via system resolver (`dns.lookup`)
- Applied to both proxy host `forward_host` and custom location `forward_host`
- If resolution fails, it safely falls back to the original hostname

Behavior:
- No behavior change unless the new flag is explicitly enabled

### 2) Allow direct PEM paste in Custom Certificate modal (issue #5347)

Problem:
Custom certificate flow previously required local file selection, which is inconvenient in remote/mobile/thin-client workflows.

What changed:
- Added textarea inputs for:
  - Certificate PEM
  - Certificate Key PEM
  - Intermediate Certificate PEM (optional)
- Kept existing file upload inputs unchanged
- Submit logic now accepts either:
  - uploaded files, or
  - pasted PEM text (converted to `File` objects client-side)
- Existing backend `validate` and `upload` multipart APIs are reused without contract changes

Behavior:
- Backward compatible with existing file upload workflow

### Docs
- Added advanced config documentation for `NPM_PRE_RESOLVE_UPSTREAM_HOSTS`

### DIFF
--- a/backend/internal/nginx.js
+++ b/backend/internal/nginx.js
@@ -1,4 +1,6 @@
 import fs from "node:fs";
+import { lookup } from "node:dns/promises";
+import { isIP } from "node:net";
 import { dirname } from "node:path";
 import { fileURLToPath } from "node:url";
 import _ from "lodash";
@@ -8,6 +10,7 @@ import { debug, nginx as logger } from "../logger.js";
 
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
+const TRUE_ENV_VALUES = new Set(["1", "true", "yes", "on"]);
 
 const internalNginx = {
 	/**
@@ -172,6 +175,8 @@ const internalNginx = {
 						locationCopy.forward_path = `/${splitted.join("/")}`;
 					}
 
+					locationCopy.forward_host = await internalNginx.preResolveUpstreamHost(locationCopy.forward_host);
+
 					renderedLocations += await renderEngine.parseAndRender(template, locationCopy);
 				}
 			};
@@ -238,10 +243,14 @@ const internalNginx = {
 				locationsPromise = Promise.resolve();
 			}
 
+			const forwardHostPromise = internalNginx.preResolveUpstreamHost(host.forward_host).then((resolvedHost) => {
+				host.forward_host = resolvedHost;
+			});
+
 			// Set the IPv6 setting for the host
 			host.ipv6 = internalNginx.ipv6Enabled();
 
-			locationsPromise.then(() => {
+			Promise.all([locationsPromise, forwardHostPromise]).then(() => {
 				renderEngine
 					.parseAndRender(template, host)
 					.then((config_text) => {
@@ -258,8 +267,8 @@ const internalNginx = {
 						reject(new errs.ConfigurationError(err.message));
 					});
 			});
-		});
-	},
+			});
+		},
 
 	/**
 	 * This generates a temporary nginx config listening on port 80 for the domain names listed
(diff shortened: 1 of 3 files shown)


jc21:
## Code Review

### Backend: `preResolveUpstreamHost` (`backend/internal/nginx.js`)

**Positives:**
- Correctly opt-in via env flag (default disabled) — zero behavior change for existing users.
- Safe fallback on DNS failure.
- `canPreResolveUpstreamHost` guard against variables (`$`), paths (`/`), ports (`:`), and literal IPs is well-considered.

**Issues:**

**1. Single-address resolution can break load-balanced upstreams (Medium)**
`dns.lookup` with `all: false` returns only one IP. If the upstream hostname resolves to multiple addresses (e.g. a service behind round-robin DNS), pre-resolution pins config to a single IP, silently degrading load balancing. Worth documenting this limitation.

**2. Config staleness in dynamic environments (Medium)**
The staleness caveat is documented, but deserves a stronger warning. In Docker environments with recreated containers, the pre-resolved IP in the static Nginx config will silently route to a dead container until the proxy host config is manually regenerated — which is likely the most common use-case for this feature.

**3. `generateConfig` mixes async patterns inconsistently (Low)**
`renderLocations` uses `await` for `preResolveUpstreamHost`, but `generateConfig` continues with `.then()` chains. The `forwardHostPromise` also mutates the passed-in `host` object as a side effect inside a `.then()`. A unified async/await approach would be cleaner and less error-prone.

**4. Port-suffixed hosts silently excluded (Low)**
`canPreResolveUpstreamHost` excludes hosts containing `:` (to skip `host:8080` style upstreams). This is correct but undocumented — worth a note since users enabling this feature might expect `host.docker.internal:8080` to be pre-resolved.

**5. No unit tests added (Low)**
`preResolveUpstreamHostEnabled`, `canPreResolveUpstreamHost`, and `preResolveUpstreamHost` are pure/near-pure functions that are straightforward to unit test. Given this touches config generation, tests would add meaningful confidence.

---

### Frontend: `CustomCertificateModal.tsx`

**Positives:**
- `buildPemFile` helper is clean and reusable.
- Reuses existing backend API contract without changes — good for backward compatibility.
- Manual validation in `onSubmit` correctly replaces the now-removed `required` HTML attribute.

**Issues:**

**1. No "OR" affordance between file input and textarea (High — UX)**
File upload and textarea are rendered one after the other with no visual separator, label, or "OR" divider. A user who provides both will silently have the file take precedence (per `buildPemFile`'s `if (file) return file` logic). This silent override should be communicated clearly.

**2. Textareas have no `<label>` elements (Medium — Accessibility)**
The three new `<textarea>` fields (`certificateText`, `certificateKeyText`, `intermediateCertificateText`) have no associated `<label>`. The file inputs above them have labels; the textareas do not. This breaks screen reader support and keyboard navigation for these fields.

**3. Noisy indentation changes in the diff (Low)**
A large portion of the diff is indentation-only changes on pre-existing `<Field>` blocks. This makes the diff harder to review and risks obscuring real changes. Ideally pre-existing code wouldn't be reformatted in the same PR.

**4. No client-side PEM format validation (Low — UX)**
The placeholder `-----BEGIN CERTIFICATE-----` is a helpful hint, but a lightweight check (e.g. trimmed content starts with `-----BEGIN`) before submit would catch obvious paste errors before hitting the backend.

---

### Documentation (`docs/src/advanced-config/index.md`)

Clean and accurate. Suggest adding:
- A note that hosts with explicit ports (e.g. `host.docker.internal:8080`) are not eligible for pre-resolution.
- A note that multi-address upstreams will be pinned to a single IP.

---

### Summary

| Area | Concern | Severity |
|---|---|---|
| Backend | Single-IP resolution breaks load-balanced upstreams | Medium |
| Backend | Config staleness warning needs strengthening in docs | Medium |
| Backend | Async pattern inconsistency in `generateConfig` | Low |
| Backend | No unit tests for new utility functions | Low |
| Frontend | No "OR" separator; silent file-over-text override | High |
| Frontend | Textareas missing `<label>` — accessibility issue | Medium |
| Frontend | Noisy indentation-only diff changes | Low |

The core approach is sound — opt-in flag with safe fallback for the backend, reusing the existing API contract for the frontend. The two main items to address before merge: **UX clarity** around file vs. paste priority in the modal, and a **docs/behaviour note** about single-IP resolution affecting load-balanced upstreams.

PR #5352, opened 2026-02-26
feat(i18n): complete French (fr) translation

MarioGervais:
Add 4 missing translation keys for certificate key type:
- certificates.key-type
- certificates.key-type-description
- certificates.key-type-ecdsa
- certificates.key-type-rsa

Also removed the obsolete `update-available` key.

### DIFF
--- a/frontend/src/locale/src/fr.json
+++ b/frontend/src/locale/src/fr.json
@@ -605,6 +605,9 @@
 	"test": {
 		"defaultMessage": "Test"
 	},
+	"update-available": {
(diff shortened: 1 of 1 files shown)


jc21:
The `update-available` key is in fact still used:
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/develop/frontend/src/components/SiteFooter.tsx#L68

jc21:
It would have been nice for you to fix your own mistake but that's for the contribution anyway!

Issue #5367, opened 2026-03-02
[Feature Request] Allow editing of DNS Provider configuration/API keys for existing SSL Certificates


lokeshkato:
### Feature Request
Currently, the application lacks the ability to modify the DNS Provider configuration once an SSL certificate has been created. If a user’s DNS API key expires, is rotated for security, or needs to be corrected, there is no mechanism within the UI to update these credentials. This creates a significant bottleneck for maintaining automated renewals via DNS-01 challenges.

### Current Behaviour
* **Static Configuration:** DNS API credentials are "locked" to the certificate upon creation.
* **Renewal Failure:** When an API key becomes invalid, renewals fail indefinitely with no way to update the underlying credential.
* **Manual Overhead:** To fix the issue, a user must delete the existing certificate and create a new one.
* **Broken Links:** Deleting the certificate forces the user to manually re-assign the new certificate to every individual Proxy Host that was previously linked to the old one.

### Requested Behaviour
* **Edit Functionality:** Add an **"Edit"** option for entries in the **Certificates** tab.
* **Credential Update:** Allow users to modify the **DNS Provider** selection and the **Provider Configuration**.
* **Persistence:** Saving these changes should update the underlying configuration for that specific Certificate ID, ensuring that existing Proxy Host assignments remain intact and valid for the next renewal attempt.

### Why this helps
* **Efficiency:** Eliminates the need to manually update dozens of Proxy Hosts just to change a single API string.
* **Security:** Encourages users to rotate their DNS API keys regularly, as the process would no longer be a destructive administrative task.
* **User Experience:** Aligns the SSL management workflow with the rest of the application (like Proxy Hosts, Streams, Redirection Hosts), where settings are expected to be editable rather than permanent.
* **Reduced Downtime:** Prevents certificates from expiring due to "unfixable" credential errors, ensuring services remain accessible and secure.

PR #5377, opened 2026-03-05
Edit feature for certification to update the DNS settings

lokeshkato:
### Summary
Currently, the application lacks the ability to modify the DNS Provider configuration once an SSL certificate has been created. If a user’s DNS API key expires, is rotated for security, or needs to be corrected, there is no mechanism within the UI to update these credentials. This creates a significant bottleneck for maintaining automated renewals via DNS-01 challenges.

### Technical Details

1. **Backend:** certificate.js – update() for LE DNS: merge meta, patch DB, then use same flow as create (disable hosts → requestLetsEncryptSslWithDnsChallenge → enable hosts + reload nginx), then refresh expires_on.
2. **Frontend:** "Edit DNS Settings" dropdown item for LE DNS certs.
3. **Locales:** All 22 locale JSON files (en, de, fr, etc.) updated with certificates.edit-dns-settings and certificates.dns.credentials-update.

### Screenshots
Screenshot 2026-03-05 at 2 50 21 PM
Screenshot 2026-03-05 at 2 50 35 PM

### Why this helps
* **Efficiency:** Eliminates the need to manually update dozens of Proxy Hosts just to change a single API string.
* **Security:** Encourages users to rotate their DNS API keys regularly, as the process would no longer be a destructive administrative task.
* **User Experience:** Aligns the SSL management workflow with the rest of the application (like Proxy Hosts, Streams, Redirection Hosts), where settings are expected to be editable rather than permanent.
* **Reduced Downtime:** Prevents certificates from expiring due to "unfixable" credential errors, ensuring services remain accessible and secure.

#5367

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -255,13 +255,81 @@ const internalCertificate = {
 			);
 		}
 
-		const savedRow = await certificateModel
+		let patchPayload = data;
+
+		// Let's Encrypt DNS: allow updating only dns_provider, dns_provider_credentials, propagation_seconds
+		const isLetsEncryptDns =
+			row.provider === "letsencrypt" &&
+			row.meta &&
+			row.meta.dns_challenge &&
+			data.meta &&
+			typeof data.meta === "object";
+
+		if (isLetsEncryptDns) {
+			const mergedMeta = { ...row.meta };
+			if (data.meta.dns_provider !== undefined) mergedMeta.dns_provider = data.meta.dns_provider;
+			if (data.meta.dns_provider_credentials !== undefined) {
+				mergedMeta.dns_provider_credentials = data.meta.dns_provider_credentials;
+			}
+			if (data.meta.propagation_seconds !== undefined) {
+				mergedMeta.propagation_seconds = data.meta.propagation_seconds;
+			}
+			patchPayload = { ...data, meta: mergedMeta };
+		}
+
+		let savedRow = await certificateModel
 			.query()
-			.patchAndFetchById(row.id, data)
+			.patchAndFetchById(row.id, patchPayload)
 			.then(utils.omitRow(omissions()));
 
+		if (isLetsEncryptDns) {
+			// Request a new Cert from LE via DNS challenge. Let the fun begin.
+
+			// 1. Find out any hosts that are using any of the hostnames in this cert
+			// 2. Disable them in nginx temporarily
+			// 3. Request cert
+			// 4. Re-instate previously disabled hosts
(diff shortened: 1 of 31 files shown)


jc21:
## Code Review

Three frontend issues worth addressing before merge:

### 1. Credentials pre-fill contradicts the `editMode` comment

`getInitialValues` in `DNSCertificateModal.tsx` explicitly populates `dnsProviderCredentials` from the existing certificate object, but the comment added to `DNSProviderFields.tsx` says _"credentials field is for new credentials only; existing credentials are never displayed."_ These are contradictory — pick one behaviour and align the code and comment. If the intent is to never show existing credentials (which the edit label "New Credentials File Content (replaces existing)" implies), the initial value should be `""`.

### 2. `useEffect` dependency fragility in `DNSProviderFields`

```ts
useEffect(() => {
  if (selectedOption && (v.meta?.dnsProviderCredentials ?? "") === "") {
    setFieldValue("meta.dnsProviderCredentials", selectedOption.credentials);
  }
}, [selectedOption, selectedOption?.credentials, v.meta?.dnsProviderCredentials, setFieldValue]);
```

`selectedOption` is derived inline via `options.find(...)` and gets a new object reference on every render. The guard condition prevents most problems, but in edit mode where credentials are pre-filled, if a user clears the field it could overwrite with the plugin template rather than leaving it blank. Using a stable ID (e.g. `dnsProviderId`) as the key dependency would be more reliable.

### 3. Modal indentation regression

The body of the `EasyModal.create` callback in `DNSCertificateModal.tsx` is indented at 8 spaces instead of the file's standard 4, introduced during the edit-mode refactor.

jc21:
CI is currently failing because of missing examples in the swagger schema:

```
14:09:45  cypress-1  | results/swagger-schema.json:2859:33  ▲           media type is missing `examples` or `example`                               
14:09:45  cypress-1  | $.paths['/nginx/certificates/{certID}'].put.requestBody.content['application/json']
14:09:45  cypress-1  | rule: oas3-missing-example  category: Examples
14:09:45  cypress-1  | 
14:09:45  cypress-1  | results/swagger-schema.json:2864:19  ▲           media type schema property `meta` is missing `examples` or `example`        
14:09:45  cypress-1  | $.paths['/nginx/certificates/{certID}'].put.requestBody.content['application/json'].schema.properties['meta']
14:09:45  cypress-1  | rule: oas3-missing-example  category: Examples
```

lokeshkato:
@jc21 , the aforementioned review comments are resolved now. Please take another look when you get a chance.

Issue #4983, opened 2025-11-23
fresh install from docker bug with capitalize letters in email


grand-lotus-iroh:
If you type in email with capital letter, you get error

"Invalid email or password"

and than it redirects and is broken and you must remove and redeploy with docker command

YuriySamorodov:
Can confirm the same issue.


**Workaround**:
In my case making both email, user name and nickmame lowercase allowed me to login successfully.
It happily accepted upper case letters, special characters and digits for password though.


**Environment**:
```
Linux 1764207385511 6.8.0-88-generic #89-Ubuntu SMP PREEMPT_DYNAMIC
Docker version 29.0.4, build 3247a5a
Nginx Proxy Manager Version 2.13.5 (847c58b) 2025-11-18 11:25:13 UTC, 
OpenResty 1.27.1.2
Certbot certbot 5.1.0
```

Bi11:
I can confirm this behavior.

It appears that PR #4664 only fixed this issue when the admin is created via the INITIAL_ADMIN_EMAIL environment variable.

However, the backend logic for creating a user via the UI (Initial Setup) was not updated. It still saves the email exactly as input (preserving uppercase), while the login validator forces lowercase. This mismatch causes the permanent lockout.

The issue seems to be located in backend/internal/user.js:
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/847c58b17081f1b6ed1d52e5b81ab0296f9f6c18/backend/internal/user.js#L26-L40

The email field in the payload needs to be normalized (e.g., .toLowerCase()) before being inserted into the database.

JeftaDirksen:
Please reopen, this is annoying as f* (since I always use capitals in my email)
Setup UI is way better than default initial credentials, but please make the casing flexible (validate username/email case insensitive).
Also it looks like PR #4665 / Commit #4664 isn't working anymore as well (INITIAL_ADMIN_EMAIL is case-sensitive again? not tested but the .toLowerCase() isn't there anymore)

nate2014jatc:
Good afternoon, just ran into this error setting up a new copy for a friend.

NameOfTheDragon:
Confirmed issue. 

Nginx Proxy Manager version: 2.15.1

Steps to reproduce:

1. Create a user using an email address containing uppercase characters, e.g.

   Test@timtek.net

2. Set a password.
3. Attempt to log in using the same email address and password.

Expected result:

Login succeeds.

[...]

Issue #5323, opened 2026-02-18
First-run admin setup fails with "Invalid email or password" and locks instance


jesperbruhn:
At first startup, I enter my information:

Name, email and password.

And press Save, but immediately get the error Invalid email or password, but am sent to the login screen, and then I can't proceed.

<img width="620" height="845" alt="Image" src="https://github.com/user-attachments/assets/f321b2c8-32a4-4252-9dd0-9aa8e3d7e52e" />

arthurbrazil:
I am also experiencing this issue.

jesperbruhn:
### [Bug] First-run admin setup fails with "Invalid email or password" and locks instance

### Description
On a fresh installation, Nginx Proxy Manager opens directly in the first-run admin setup wizard.  
After entering name, email and password and pressing **Save**, the UI returns **"Invalid email or password"** and redirects to the login screen.  
The instance becomes locked because no credentials work afterwards.

It appears the admin record is partially written to the SQLite database, leaving it in an inconsistent state.

### Expected behaviour
- Setup wizard creates the admin account successfully
- User is redirected to the dashboard

### Actual behaviour
- Error message appears after pressing Save
- Redirect to login page
- Login fails with the newly created credentials

### Steps to reproduce
1. Fresh install of Nginx Proxy Manager
2. Open web UI (setup wizard appears)
3. Enter name, email and password
4. Click Save
5. Observe error and failed login

### Environment
- Installation method: Docker + Docker Compose
- Host: Proxmox LXC (unprivileged)
- OS: Debian 12
- Database: SQLite (default)
- Browser: Chrome (also tested Incognito)

### Workaround
docker compose down
sqlite3 /opt/npm/data/database.sqlite "UPDATE user SET is_deleted=1;"
docker compose up -d

### Notes
- Using lowercase email and avoiding browser autofill prevented the issue after reset.
- Possibly related to how the initial admin user is written during setup.

IDEDARY:
I am also experiencing this issue. Not the greatest first impression with this software to be honest.

Momentum6890:
Same issue here

Zum0shi:
Same here, @jesperbruhn thank you for the post, the workaround fixed it for me!

Nividica:
Also ran into this issue, but was able to work around it by starting fresh and creating the new user as
- Email: `admin@example.com`
- Password: `changeme`

Then changing the email and password once inside.

[...]

PR #5442, opened 2026-03-30
Ensure email is lowercase and check availability on user creation

ChinoKou:
This pull request adds email normalization and uniqueness validation to the user creation logic in `backend/internal/user.js`. Now, emails are converted to lowercase and trimmed, and the system checks if the email is already in use before proceeding.

**User validation improvements:**

* Emails are now normalized (lowercased and trimmed) before processing to ensure consistency.
* Added a uniqueness check for emails during user creation, throwing a validation error if the email is already in use.


ChinoKou:
Fixes: #4983 #5323

jc21:
## Code Review

Thanks for this PR — closing the gap between `create` and `update` is the right thing to do. The `update` path already had both of these checks, so it's good to bring `create` in line.

A couple of issues worth addressing before merging:

---

### 1. Missing null/undefined guard (medium)

```js
data.email = data.email.toLowerCase().trim();
```

If `data.email` is `null` or `undefined`, this throws an unhandled `TypeError` rather than a clean validation error. The `update` path guards against this first:

```js
if (typeof data.email !== "undefined") {
    data.email = data.email.toLowerCase().trim();
```

Even if upstream schema validation currently guarantees `email` is present, matching this pattern would be more defensive and consistent.

---

### 2. Uniqueness check runs before access control (low)

The new block is inserted before `access.can("users:create", data)`. This means an unauthenticated or unauthorized caller can probe whether an email is already in use before being rejected for lack of permissions. The check should move to after the access check.

---

### 3. Explicit `undefined` argument (nit)

```js
await internalUser.isEmailAvailable(data.email, undefined);
```

Passing `undefined` explicitly adds noise — omitting the second argument or passing `null` would be cleaner.

---

### Suggested ordering

```js
await access.can("users:create", data); // access check first

if (typeof data.email !== "undefined" && data.email !== null) {
    data.email = data.email.toLowerCase().trim();
    const emailAvailable = await internalUser.isEmailAvailable(data.email);
    if (!emailAvailable) {
        throw new errs.ValidationError(`Email address already in use - ${data.email}`);
    }
}

data.avatar = gravatar.url(data.email, { default: "mm" });
```

The error message format and the overall approach are consistent with the existing `update` logic — just the ordering and guard need a tweak.

ChinoKou:
> ## Code Review
> Thanks for this PR — closing the gap between `create` and `update` is the right thing to do. The `update` path already had both of these checks, so it's good to bring `create` in line.
> 
> A couple of issues worth addressing before merging:
> 
> ### 1. Missing null/undefined guard (medium)
> ```js
> data.email = data.email.toLowerCase().trim();
> ```
> 
> If `data.email` is `null` or `undefined`, this throws an unhandled `TypeError` rather than a clean validation error. The `update` path guards against this first:
> 
> ```js
> if (typeof data.email !== "undefined") {
>     data.email = data.email.toLowerCase().trim();
> ```
> 
> Even if upstream schema validation currently guarantees `email` is present, matching this pattern would be more defensive and consistent.
> 
> ### 2. Uniqueness check runs before access control (low)
> The new block is inserted before `access.can("users:create", data)`. This means an unauthenticated or unauthorized caller can probe whether an email is already in use before being rejected for lack of permissions. The check should move to after the access check.
> 
> ### 3. Explicit `undefined` argument (nit)
> ```js
> await internalUser.isEmailAvailable(data.email, undefined);
> ```
> 
> Passing `undefined` explicitly adds noise — omitting the second argument or passing `null` would be cleaner.
> 
> ### Suggested ordering
> ```js
> await access.can("users:create", data); // access check first
> 
> if (typeof data.email !== "undefined" && data.email !== null) {
>     data.email = data.email.toLowerCase().trim();
>     const emailAvailable = await internalUser.isEmailAvailable(data.email);
>     if (!emailAvailable) {
>         throw new errs.ValidationError(`Email address already in use - ${data.email}`);
>     }
> }
> 
> data.avatar = gravatar.url(data.email, { default: "mm" });
> ```
> 
> The error message format and the overall approach are consistent with the existing `update` logic — just the ordering and guard need a tweak.

Thanks for the great feedback! I've pushed the updates to fix the ordering, add the null guard, and clean up the method call.

Also, spot on about the validation. The frontend already checks this, but I completely agree the backend check is necessary. It adds a solid layer of defense-in-depth for direct API calls.

PR #5455, opened 2026-04-05
i18n(zh-TW): Add Traditional chinese 

neko0xff:
- This fix issue:
  * Add New Translation
  * ISO code 
    1. zh-TW: Traditional Chinese (Taiwan)
    2. zh-CN: Simplified Chinese (China)


jc21:
Thanks for the contribution. Please fix the conflicts and remove the lock files from the PR as they should only be committed when modifying node packages.

joengan-evertoprich:
Thanks for adding Traditional Chinese support.

Would it be more standards-aligned to use `zh-Hant` instead of `zh-TW`,
and rename the existing Simplified Chinese locale to `zh-Hans` instead
of `zh-CN`?

Under BCP 47 / RFC 5646:

- `Hant` and `Hans` are script subtags, representing Traditional and
  Simplified Chinese writing systems.
- `TW`, `HK`, `MO`, and `CN` are region subtags.

If this translation is intended to be a general Traditional Chinese
translation shared by users in Taiwan, Hong Kong, Macau, and elsewhere,
`zh-Hant` would describe it more accurately than `zh-TW`.

Likewise, `zh-Hans` would be more accurate than `zh-CN` if the existing
translation is intended to represent Simplified Chinese generally,
rather than specifically Mainland Chinese localization.

Region-specific locales such as `zh-Hant-TW` or `zh-Hant-HK` could still
be added later if different terminology, date formats, or regional
conventions are needed.

References:
- BCP 47 / RFC 5646
- W3C guidance on choosing language tags
- Unicode CLDR locale conventions

PR #5461, opened 2026-04-06
Performance: AJV schema compiled on every validation call (api.js)

barttran2k:
## Problem

In `backend/lib/validator/api.js`, `ajv.compile(schema)` is called inside `apiValidator` on every single request. AJV schema compilation is computationally expensive and should be done once per schema, then the compiled validate function should be reused. Under load, this causes significant CPU overhead on every API request that involves validation.

**Severity**: `high`
**File**: `backend/lib/validator/api.js`

## Solution

Cache compiled validators using a Map keyed by schema (or schema $id). For example: `const cache = new Map(); function getValidator(schema) { let v = cache.get(schema); if (!v) { v = ajv.compile(schema); cache.set(schema, v); } return v; }` Then call `getValidator(schema)` instead of `ajv.compile(schema)` on each invocation.

## Changes

- `backend/lib/validator/api.js` (modified)
- `backend/lib/validator/index.js` (modified)

## Testing

- [ ] Existing tests pass
- [ ] Manual review completed
- [ ] No new warnings/errors introduced

### DIFF
--- a/backend/lib/validator/api.js
+++ b/backend/lib/validator/api.js
@@ -9,6 +9,8 @@ const ajv = new Ajv({
 	coerceTypes: true,
 });
 
+const compiledSchemas = new WeakMap();
+
 /**
  * @param {Object} schema
  * @param {Object} payload
@@ -25,7 +27,11 @@ const apiValidator = async (schema, payload /*, description*/) => {
 	}
 
 
-	const validate = ajv.compile(schema);
+	let validate = compiledSchemas.get(schema);
+	if (!validate) {
+		validate = ajv.compile(schema);
+		compiledSchemas.set(schema, validate);
+	}
 
 	const valid = validate(payload);
 
--- a/backend/lib/validator/index.js
+++ b/backend/lib/validator/index.js
@@ -14,6 +14,8 @@ const ajv = new Ajv({
 	schemas: [commonDefinitions],
 });
 
+const compiledSchemas = new WeakMap();
+
 /**
  *
  * @param   {Object} schema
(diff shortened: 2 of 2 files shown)


jc21:
## Code Review

Thanks for the contribution! The intent here is valid — avoiding repeated AJV schema compilation is a real concern — but the two files need to be assessed separately.

### `api.js` — Works correctly ✅

The cache works as intended here. `getValidationSchema()` returns a property reference from the module-level `compiledSchema` singleton (cached in `backend/schema/index.js`), so the same object reference is returned on every request for the same route. `WeakMap.get(schema)` will hit after the first call.

**Why it helps**: AJV 8 internally caches compiled schemas using `stableStringify(schema)` to generate a lookup key. Even with AJV's own caching, `stableStringify` runs on every `ajv.compile()` call. The `WeakMap` lookup by object reference avoids that serialization cost on subsequent requests.

### `index.js` — Cache never hits ❌

This is the critical issue. The `validator` exported from `index.js` is called throughout the codebase with **inline schema object literals**:

```js
// e.g. backend/routes/nginx/proxy_hosts.js
const data = await validator(
    {
        additionalProperties: false,
        properties: {
            expand: { $ref: "common#/properties/expand" },
        },
    },
    req.body
);
```

Each invocation of the route handler creates a **new object in memory**. `WeakMap` uses object identity (reference equality), so `compiledSchemas.get(schema)` will always return `undefined` — the cache never hits. `ajv.compile(schema)` is called on every request exactly as before, making the added code dead weight.

**Fix options:**
1. Move the inline schema objects in the callers to module-level constants so the same reference is reused across requests — then the `WeakMap` approach works.
2. Use a string-keyed `Map` with a deterministic key (e.g. an assigned `$id` on each schema).
3. Use AJV's own `ajv.addSchema(schema, key)` / `ajv.getSchema(key)` APIs with explicit string keys.

### Other notes

- The "high severity" label is overstated. This is a proxy manager admin panel with low request throughput, AJV already does its own internal caching, and schemas for `api.js` are already long-lived module-level objects.
- No tests are included or checked off in the PR template.

Happy to see a revision that fixes the `index.js` callers to hoist their schemas to module-level constants — that would make both changes effective.

PR #5479, opened 2026-04-13
Add dependency for DNSPod plugin

### LINKED ISSUE #2440: ModuleNotFoundError: No module named 'zope'
When I use the SSL function with the latest version, the software reports an error：“ModuleNotFoundError: No module named 'zope'”。
When I use the same configuration and another version of github-pr-2411 to start the ssl function, there is no problem.

Mchnan:
Because DNSPod plugin on pypi is very old, it still needs zope-interface to work. Fixes #2440

### DIFF
--- a/backend/certbot/dns-plugins.json
+++ b/backend/certbot/dns-plugins.json
@@ -177,7 +177,7 @@
 	},
 	"dnspod": {
 		"credentials": "dns_dnspod_email = \"email@example.com\"\ndns_dnspod_api_token = \"id,key\"",
-		"dependencies": "",
+		"dependencies": "zope.interface==8.3",
 		"full_plugin_name": "dns-dnspod",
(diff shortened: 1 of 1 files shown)


jc21:
Can you please comment and verify that the built image fixes the issue?

Mchnan:
This **PR** successfully fixes the **DNSPod** plugin.

However, after double-checking similar proxy projects, I noticed they tend to deprecate this specific plugin in favor of the **tencentcloud** plugin.

**Additional Context:**
This plugin is heavily used within Mainland China. Unfortunately, a vast majority of Chinese tutorials still point to DNSPod, making many users (including myself) fall into this legacy trap. Breaking this inertia will be difficult in the short term without a major ecosystem shift. Therefore, maintaining this fix remains necessary for users relying on those older guides.

PR #5506, opened 2026-04-24
PROXY Protocol Round 3?

reversefold:
I decided to take a look at https://github.com/NginxProxyManager/nginx-proxy-manager/pull/4505 and see what it would take to make it work. I merged adrum's branch onto develop and fixed the issues I found.

I believe this is a fully working version of the PROXY protocol support, with support for Proxy Hosts. I was able to create and edit both proxy hosts and streams in a dev environment with no issues.

I have support for PROXY protocol with streams working in local testing but it fails to pass the CI tests for reasons I don't currently understand.

Thanks to @adrum and @SBado for their prior work on this.

PR #5511, opened 2026-04-29
PROXY protocol support for Streams

reversefold:
Based on https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5506, opening as a separate PR as the stream support has been problematic in CI testing but works fine when testing locally. I am opening a draft PR to try to work out the CI issues.

If someone could point to how to run the Cypress tests locally I would happily run them locally instead of waiting for the Jenkins CI.


reversefold:
*Finally*. I found the last few things causing problems. A documented way to run the tests locally would have made this much quicker to debug, not to mention CI which actually flags what exactly is failing.

jc21:
## Code Review

Thanks for the PR — PROXY protocol support is a useful addition. Here are some issues I found that should be addressed before merge.

---

### Blocking Bugs

**1. Migration `down()` uses wrong variable and wrong knex parameter — will crash on rollback**

Both migration files define `migrate_name` (snake_case) at the top but reference `migrateName` (camelCase) inside `down()`, which is `undefined` and will throw a `ReferenceError`. Additionally, the `down()` parameter is named `_knex` but the function body calls `knex.schema` — a second crash on the same path.

Affected files:
- `backend/migrations/20220209144645_proxy_protocol.js` (lines 48–57)
- `backend/migrations/20260425014442_proxy_protocol_stream.js` (lines 95–104)

```js
// Both files have this double bug:
const down = (_knex) => {                          // parameter is _knex...
    logger.info(`[${migrateName}] Migrating Down...`); // but uses migrateName (undefined)
    return knex.schema.table(...)                  // and uses knex (also undefined)
```

---

**2. Migration filename is dated 2022 — incorrect run order**

`20220209144645_proxy_protocol.js` will be ordered before all migrations that have run since February 2022, which could cause schema conflicts. It should be renamed with the current date (e.g. `20260514XXXXXX_proxy_protocol.js`).

---

**3. Trailing comma in JSON schema files — will cause parse errors**

JSON does not permit trailing commas. These two files have one after the `enable_proxy_protocol` `$ref`:

- `backend/schema/paths/nginx/streams/post.json`
- `backend/schema/paths/nginx/streams/streamID/put.json`

```json
"enable_proxy_protocol": {
    "$ref": "...#/properties/enable_proxy_protocol",  // ← trailing comma
},
```

This will break schema validation at runtime.

---

**4. Debug comment rendered into every stream nginx config**

`backend/templates/stream.conf` contains:

```
# {{ enable_proxy_protocol }}
```

This renders a nginx comment with the raw field value into every generated stream config. It needs to be removed.

---

### Bugs

**5. `stream.conf` boolean check is incomplete**

`stream.conf` only checks `enable_proxy_protocol == 1`, not `== true`. All other templates in this PR check both conditions:

```liquid
{%- if enable_proxy_protocol == 1 or enable_proxy_protocol == true %}
```

If the value comes through as a boolean `true` rather than integer `1`, proxy protocol will silently not activate for streams.

---

**6. `real_ip_header proxy_protocol` emitted without `set_real_ip_from`**

In `backend/templates/_proxy_protocol.conf`:

```liquid
{% if load_balancer_ip != '' %}
  set_real_ip_from {{ load_balancer_ip }};
{% endif %}
real_ip_header proxy_protocol;
```

If `load_balancer_ip` is empty, `real_ip_header proxy_protocol;` is still written to the config without any `set_real_ip_from` directive. nginx requires at least one `set_real_ip_from` for `real_ip_header` to function — without it nginx may log warnings and not extract the real IP. Either make `load_balancer_ip` required when proxy protocol is enabled, or guard `real_ip_header proxy_protocol;` inside the same `if` block.

---

### Minor

**7. Inconsistent UI between modals**

`ProxyHostModal.tsx` applies a `bg-lime` class to the toggle when proxy protocol is enabled; `StreamModal.tsx` does not. The two toggle implementations also diverge — one spreads `{...field}`, the other manually wires `checked` and `onChange`. Worth making consistent.

**8. `loadBalancerIp` validation not wired into Formik**

The `loadBalancerIp` input sets `required={form.values.enableProxyProtocol}` but there is no corresponding Formik `validate` function or Yup schema entry. The HTML `required` attribute alone won't block form submission.

**9. Dead code comments in JSX**

`ProxyHostModal.tsx` and `StreamModal.tsx` both contain large commented-out blocks of legacy AngularJS-style markup. These should be removed rather than left as dead code.

---

### Test Coverage

The Cypress tests are appropriately extended to assert `enable_proxy_protocol` defaults to `false`. A test that actually enables the flag and verifies the generated nginx config content would be valuable once CI is stable.

jc21:
> _Finally_. I found the last few things causing problems. A documented way to run the tests locally would have made this much quicker to debug, not to mention CI which actually flags what exactly is failing.

CI logs are public, for this PR you can find your builds here: https://ci.nginxproxymanager.com/job/nginx-proxy-manager/job/PR-5511/ This is also linked in the `Checks` on this PR.

Also, since this proxy change needs ports 88/444, you may want to  consider adding documentation to `docs/src/advanced-config/index.md` mentioning that they'll need to expose those ports in their docker configuration

reversefold:
Pushed fixes for 1-6 & 9, as well as an update to the advanced config docs.

For comment number 7, I was following the styles of the files that I was editing. The reason that the proxy host checkbox uses lime for the color is because the rest of the toggles in that file use it. The reason that the stream form does not use lime is because the toggles there do not. Same answer for spreading of `{...field}`, the proxy host model uses that, the stream model does not. Bringing those two files into parity is not within the scope of this PR.

For comment number 8, I cannot submit the form with no value in the "load balancer ip" field, so it seems to be working as intended from a UX perspective, at least as far as a value being required. Could you give some pointers as to where and how the YUP schema or validation would be implemented? I am not familiar with Formik and was working primarily off of examples elsewhere in this project.

PR #5515, opened 2026-05-02
Changed filenames to include first domain or inbound port

### LINKED ISSUE #746: Change access log filenames to domain name (or add option)

joewesch:
Closes: #746

This PR is an upgraded replacement for #2606.

This PR changes all config filenames to be `id.first_domain` for proxy_host, redirection_host or dead_host and `id.incoming_port` for stream. Each type will fall back to just `id` if there are no domains or if the type is undefined. All log files will also be renamed accordingly automatically.

```bash
# Before
[root@docker-a8439ada89f2:/app]# ls /data/nginx/proxy_host/
1.conf
[root@docker-a8439ada89f2:/app]# ls /data/logs/
fallback_error.log
fallback_http_access.log
fallback_http_error.log
fallback_stream_access.log
proxy-host-1_access.log
proxy-host-1_error.log

# After
[root@docker-a8439ada89f2:/app]# ls /data/nginx/proxy_host/
1.test.example.com.conf
[root@docker-a8439ada89f2:/app]# ls /data/logs/
fallback_error.log
fallback_http_access.log
fallback_http_error.log
fallback_stream_access.log
proxy-host-1.test.example.com_access.log
proxy-host-1.test.example.com_error.log
```

This will also happen if the user either adds a domain to the host that sorts alphabetically first or if they remove the first domain. These changes will not take effect immediately, which seems to be in-line with the rest of the upgrades in this project. They will instead update the next time each host is saved.


jc21:
Thanks for the contribution. I've asked Claude for a review and here's the parts I agree with:

-----------

### Issues

1. Bug: `letsencryptTempConfigFilenamesToDelete` only deletes `letsencrypt_{id}.{slug}.conf`, not `letsencrypt_{id}.conf`

```
// backend/lib/utils.js
const prefix = `letsencrypt_${certificateId}.`;
return dirEntries.filter((name) => name.startsWith(prefix) && name.endsWith(".conf"));
```
`"letsencrypt_3.conf".startsWith("letsencrypt_3.")` is actually true (the .conf follows the .), so this does work. However, the test for this function doesn't exercise the bare `letsencrypt_{id}.conf` form with `deleteErrFile=false/true`, so the coverage is incomplete. Worth adding.

2. Mutation side-effect on host object

```
// backend/internal/nginx.js:122
host.nginx_file_stem = utils.getNginxFileStem(nice_host_type, host);
```
`generateConfig` mutates the caller's object to inject nginx_file_stem for template rendering. This is a hidden side-effect — callers don't expect their host object to gain a new property. The stem could instead be passed as a separate template variable rather than attaching it to host.

backend/internal/nginx.js
```diff
@@ -196,7 +244,10 @@ const internalNginx = {
 
 		return new Promise((resolve, reject) => {
 			let template = null;
-			const filename = internalNginx.getConfigName(nice_host_type, host.id);
+
+			host.nginx_file_stem = utils.getNginxFileStem(nice_host_type, host);
```
joewesch, line 250: > `generateConfig` mutates the caller's object to inject nginx_file_stem for template rendering.

This is actually incorrect. The `host` constant was already created as a clone (deep copy) of the `host_row` from the caller just a few lines above to prevent this exact situation (hence the comment):

https://github.com/NginxProxyManager/nginx-proxy-manager/blob/9875a7650d33254bad06273a355cd77517f5ffa8/backend/internal/nginx.js#L188-L190

backend/lib/utils.test.js
```diff
@@ -0,0 +1,112 @@
...
+
+describe("letsencryptTempConfigFilenamesToDelete", () => {
+	it("selects letsencrypt_id.conf and letsencrypt_id.slug.conf", () => {
+		const names = ["letsencrypt_3.conf", "letsencrypt_3.app.example.conf", "letsencrypt_4.conf"];
+		assert.deepEqual(letsencryptTempConfigFilenamesToDelete(names, 3).sort(), [
+			"letsencrypt_3.app.example.conf",
+			"letsencrypt_3.conf",
+		]);
+	});
+});
```
joewesch, line 112: > Bug: `letsencryptTempConfigFilenamesToDelete` only deletes `letsencrypt_{id}.{slug}.conf`, not `letsencrypt_{id}.conf`
> 
> However, the test for this function doesn't exercise the bare `letsencrypt_{id}.conf` form with `deleteErrFile=false/true`, so the coverage is incomplete. Worth adding.

Claude seems to be hallucinating here, I think? The `deleteErrFile` param is part of `diskConfigFilenamesToDelete`, not `letsencryptTempConfigFilenamesToDelete`. Regardless, I explicitly test the bare ID in **both** cases. I have `1.conf` on line 94 and `letsencrypt_3.conf` on line 106.

joewesch:
@jc21 is there anything still pending for this?

Haswell-s:
@joewesch Any updates?

joewesch:
@Haswell-s I don't have any updates. I'm still waiting for someone to approve and merge this. I have provided all I can.

Issue #3158, opened 2023-08-25
Table improvements: Sorting on columns


khalidzamer:
**Is your feature request related to a problem? Please describe.**
Improvements to tables :
- Ability to sort order in the table view

**Describe the solution you'd like**
When the list of hosts/certificates gets longer, I would like to be able to sort order, for example, sort by source alphabetically, sort by status, created on etc..  

**Describe alternatives you've considered**
I temporarily copy the data to another tool with sortable columns and organize there, but its not practical

**Additional context**
Could we also add `created` as a column instead of subtext, or introduce a `select columns` feature to adjust the table columns in addition to the sort ability

Rowanooijen:
+1

iluapot:
+1

iluapot:
Is there an open duplicate for this or has a decision been made? Otherwise, please keep it open, it'd be a good enhancement.

PR #5520, opened 2026-05-05
Adds column sorting to Nginx tables

clementfavre:
Issue #3158 

Adds sorting on the columns of the Nginx tables (Proxy Hosts, Redirection Hosts, Dead Hosts and Streams)
The project already uses TanStack React Table so it was mostly a matter of enabling `getSortedRowModel` and adding a small sort indicator in the table header

Sortable columns
- Proxy Hosts => Source, Destination, Status
- Redirection Hosts => Source, HTTP Code, Scheme, Destination, Status
- Dead Hosts => Source, Status
- Streams => Incoming Port, Destination, Status


jc21:
Great work, thanks :)

Issue #5477, opened 2026-04-12
Feature Request: Customizable Instance Name / Page Title


dont-touchme:
### Problem

When managing multiple Nginx Proxy Manager instances simultaneously (e.g., production, staging, home lab), all instances share the same browser tab title **"Nginx Proxy Manager"**. This makes it difficult to distinguish between tabs at a glance, increasing the risk of accidentally applying changes to the wrong instance.

### Proposed Solution

Add an **Instance Name** setting in the Admin UI (under Settings), which would:

- Replace or append to the browser `<title>` tag (e.g., `"Nginx Proxy Manager - Production"`)
- Optionally display the instance name in the navigation bar or header
- Be configurable via both the UI and an environment variable (e.g., `INSTANCE_NAME`) for Docker-first workflows

**Example environment variable usage:**
```yaml
environment:
  INSTANCE_NAME: "Production"
```

**Example resulting title:**
```
Nginx Proxy Manager - Production
```

### Use Case

Users managing 2+ NPM instances open in separate browser tabs have no visual way to tell them apart without hovering over each tab or reading the URL. A simple customizable name solves this immediately.

### Alternatives Considered

- Manually patching frontend files inside the container — breaks on every update
- Using a reverse proxy with `sub_filter` — adds unnecessary infrastructure complexity
- Browser userscripts (Tampermonkey) — requires per-user setup, not a proper solution

### Additional Context

This is a commonly requested quality-of-life improvement for homelab and production users running multiple NPM instances. Similar tools like Portainer and Grafana already support custom instance naming.

PR #5526, opened 2026-05-08
Feature Request: Customizable Instance Name / Page Title

TomicaSoft:
#5477
Docker environment INSTANCE_NAME to add title suffix in window title and on the main page title.

`
docker run --rm -p 81:81 -v data:/data -v letsencrypt:/etc/letsencrypt -e INSTANCE_NAME="My Prod" nginxmanager:latest
`

image

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -39,6 +39,9 @@ EXPOSE 80 81 443
 COPY backend       /app
 COPY frontend/dist /app/frontend
 
+COPY docker/scripts/entrypoint /entrypoint
+RUN chmod +x /entrypoint
+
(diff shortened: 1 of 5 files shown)


jc21:
Is there any reason this couldn't be part of the "Settings" section of the app instead of in a Docker env var?

TomicaSoft:
It was one of the wishes in the feature request and implementing it this way was simple for me.
I did it so that you can change it at any time in Docker without build own image and I think this configuration is not something that needs to be changed regularly.
You set an instance name once and don’t change it every week or so.
If you don’t like this approach, feel free to ignore the pull request.
:-)

PR #5535, opened 2026-05-13
chore(deps): bump EuroDNS plugin and fix endpoint url

DaanSelen:


DaanSelen:
Perhaps only the CI/CD Pipeline is interesting since the dependencies are getting bumped in the latest commits.

DaanSelen:
Okay... this might be more fitting.

jc21:
Thanks for the contribution, but I do not want to use github workflows at all. CI/CD is hosted at [ci.nginxproxymanager.com](https://ci.nginxproxymanager.com).

Please remove it and keep the PR focussed on a single problem; updating the dns plugin. Thanks!

DaanSelen:
> Thanks for the contribution, but I do not want to use github workflows at all. CI/CD is hosted at [ci.nginxproxymanager.com](https://ci.nginxproxymanager.com).
> 
> Please remove it and keep the PR focussed on a single problem; updating the dns plugin. Thanks!

Reverted the CI/CD and the Dependabot changes. What about the lockfiles and year increment?

backend/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "nginx-proxy-manager",
-	"version": "2.0.0",
+	"version": "2.14.1",
```
jc21: This is unnessessary, CI/CD will update this value at deployment time using the `.version` file in the root of this project.

docker/Dockerfile
```diff
@@ -3,8 +3,9 @@
 
 # This file assumes that the frontend has been built using ./scripts/frontend-build
 
+ARG BASE_IMAGE=nginxproxymanager/nginx-full:certbot-node
 FROM nginxproxymanager/testca AS testca
-FROM nginxproxymanager/nginx-full:certbot-node
+FROM $BASE_IMAGE
```
jc21, line 8: While this is technically OK, just want to point out that Docker will not used cached built layers after seeing an `ARG` line in the Dockerfile. Not a big deal for me an my CI which operates in a clean room.

frontend/package.json
```diff
@@ -1,6 +1,6 @@
 {
 	"name": "nginx-proxy-manager",
-	"version": "2.0.0",
+	"version": "2.14.1",
```
jc21: Same version logic as above

jc21:
> Reverted the CI/CD and the Dependabot changes. What about the lockfiles and year increment?

The year is fine, the package.json and lock files aren't required. I just did a full dependancy update yesterday and Dependabot made a lot of sub-dependency changes too.

frontend/package.json
DaanSelen: Not to be blunt, but is there a downside to having it roughly pinned? Curious if anything...

docker/Dockerfile
DaanSelen: My builds also happen in CI CD (although Github/Gitea/Act) so for me it allowed me to create custom based images (for when I bumped nginx-full to Debian 13 for example)

backend/package.json
DaanSelen: I will remove this.

DaanSelen:
> The year is fine, the package.json and lock files aren't required. I just did a full dependancy update yesterday and Dependabot made a lot of sub-dependency changes too.
> 

Most of the changes came from me doing it all literally a couple days before you did it. If you want I can revert the changes.

DaanSelen:
I don't know when you get the time to recheck this PR but I am going to completely rebuild it and force push it in this pr.

All to apply the feedback.

DaanSelen:
@jc21 I force pushed to what might be relevant.

DaanSelen:
@jc21 is anything more needed?

DaanSelen:
Thank you!

PR #5333, opened 2026-02-20
feat(access-list): add optional note field to access list clients

TheMazeIsAmazing:
Based on my earlier Feature Request, [nr. 5277](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/5277), I thought I might take a jab at it myself.

This PR introduces an optional “note” field for access list client rules to improve clarity when managing multiple IP entries. It allows users to annotate each rule (e.g. “Home”, “Office”, “VPN”) without affecting nginx behaviour.

Backend changes:

* Added optional `note` property to access list client schema.
* Updated access list create/update endpoints to accept and persist `note`.
* Kept `note` fully optional and backwards-compatible with existing data.
* Ensured the nginx render filter outputs the note as a comment when present, preserving valid nginx syntax. Example image:

image

Frontend changes:

* Extended `AccessListClient` type with `note?: string`.
* Added a note input field in the Access List modal (Rules tab).
* Included `note` in API payload only when non-empty.
* Added i18n placeholder text for the new field.
* Widened the Access List modal to improve layout and usability with the additional column.

 

If something is broken, or needs to be fixed, let me know. Thanks for your time!

Issue #5277, opened 2026-02-05
Feature request: Adding a note / description to each IP within the Access List


TheMazeIsAmazing:
**Is your feature request related to a problem? Please describe.**
We use NPM as a small company for easy management of hosted company tools. While currently it's quite doable to manage each colleague's home IP, I foresee when numbers grow within the company, it won't be as easy.

**Describe the solution you'd like**
That's why I think it would be great if we could add a note to each IP, specifying whose IP it is.

**Describe alternatives you've considered**
As headcounts grow, we also could use a VPN service to the office instead of whitelisting colleagues home IP. This would therefore make it, so only one IP needs to be within the Access List. 

**Additional context**
I've created the following Mock-up image:
<img width="548" height="478" alt="Image" src="https://github.com/user-attachments/assets/287f9be9-648d-4af8-b10a-606361ff33ad" />

miguelcouteirorodrigues:
I'd like to suggest extending this to other pages as well. As of right now, I have multiple streams that I'd like to add some sort of description to in order to know what each one is associated with.

PR #5540, opened 2026-05-14
feat(access-list): add optional note field to access list clients

TheMazeIsAmazing:
## Why
This PR introduces an optional “note” field for access list client rules to improve clarity when managing multiple IP entries. It allows users to annotate each rule (e.g. “Home” or “Office”) without affecting current behaviour.
(See feat-req: [https://github.com/NginxProxyManager/nginx-proxy-manager/issues/5277](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/5277))

Front-end change:
Screenshot 2026-05-14 at 14 03 49

Note added to the generated Nginx files (e.g. `/data/nginx/proxy_host/X.conf`):
image

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

## Misc.
I'm sorry, this is a new version of [https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5333](https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5333). Wanted to make sure it was up-to-date with current `develop` branch and resolve merge conflicts. Guess I did it wrong and accidentally removed my PR 🥲.


jc21:
## Code Review

Thanks for the PR and for keeping it up-to-date with `develop`. The feature is well-scoped and non-breaking — the nginx sanitization is correct and the localization coverage is thorough.

### Blocking Issue

**Global CSS override in `frontend/src/App.css`**

```css
.modal-dialog { 
    max-width: 700px !important; 
}
```

This widens **every modal** in the application, not just the Access List modal. Bootstrap's default is 500px. This needs to be scoped to only the Access List modal — e.g. via a dedicated CSS class applied through `modalClassName` or a wrapper class — to avoid a regression across all other dialogs.

---

### Minor Issues

1. **Accessibility**: The note `<input>` has no associated `<label>`, only a placeholder. Placeholders disappear on focus and are not a substitute for labels — a visually hidden label or `aria-label` would fix this.

2. **Note not visible in read view**: The note only appears during edit. Users can't see annotations without opening the edit dialog or inspecting the nginx config file. Worth considering showing the note inline in the client list within the modal.

3. **Migration `down` is a no-op**: The rollback function just logs a warning rather than dropping the column. This is an acceptable choice (dropping a column would destroy data), but it's worth a comment in the code marking it as intentional.

4. **Unrelated locale additions**: Several locales (`de`, `it`, `ja`, `pl`, `vi`, `zh`) also add `access-list.rule-source.placeholder` (the IP address placeholder), which was previously missing from those files. Worth a quick sanity-check that those IP placeholder strings are correct since they weren't the original focus of the PR.

---

### What's Good

- `client.note || null` correctly coerces empty string to `null` before persisting
- The nginx comment sanitization (`[\r\n#;]` → space) is correct — stripping newlines is the key safety measure
- Frontend only includes `note` if non-empty after trimming (in `AccessListModal.tsx`)
- Schema `maxLength: 200` matches the DB column size

Fix the global CSS scope and this is good to go.

frontend/src/modals/AccessListModal.tsx
jc21: You can apply sizes and other properties to this modal that won't affect all models. ie
```
size="lg"
```
see https://react-bootstrap.netlify.app/docs/components/modal/#optional-sizes

TheMazeIsAmazing: Just fixed it!

TheMazeIsAmazing:
Hey @jc21,

I implemented the feedback (https://github.com/NginxProxyManager/nginx-proxy-manager/pull/5540/commits/6fadcd084de2b7f1fb19d8d4b607a71985da8881), so with any luck it also could be included in v2.15 🤞? 

Thanks for your time and effort 😁

PR #5587, opened 2026-05-26
Add HTTP/3 opt-in support with migrations, tests, and cleanup

### LINKED ISSUE #1550: HTTP3 + QUIC support

ricolehn:
# feat: add HTTP/3 (QUIC over UDP) opt-in support for proxy hosts 🚀

## 📖 Description

This PR implements full, opt-in support for **HTTP/3 (QUIC over UDP)** for proxy hosts. It covers the complete stack, including backend database schema migrations, template listener generations, dynamic Nginx capability detection, frontend toggle UI with multi-language localizations, and E2E integration tests.

The feature is designed from the ground up as a **safe, zero-regression opt-in**, ensuring that existing users upgrading their installations will experience exactly zero changes or configuration risk by default.

---

## 🔗 Fixes & Resolved Issues

This implementation natively satisfies and closes the long-standing community feature requests:

Fixes #1550

* **Closes #1550** — Introduces native HTTP/3 / QUIC support into the proxy engine generation layer.
* **Eliminates UI Overwrite Collisions** — Resolves configuration loss for users attempting manual file overrides by treating QUIC as a first-class citizen inside the core orchestration engine.

---

### 1. Robust Startup & Compatibility Protection (Non-QUIC Nginx Builds)
* **The Challenge:** Nginx will refuse to boot if `quic` directives are present in configuration blocks on older binaries lacking the `--with-http_v3_module` compilation engine.
* **The Solution:** Added a synchronous startup capability check in `55-http3-check.sh` sourced inside the s6-overlay prepare oneshot service pipeline. If Nginx lacks HTTP/3 capabilities OR if `NPM_HTTP3_DISABLED=1` is set, the startup script dynamically strips `quic` listen parameters from the static `default.conf` before Nginx launches, ensuring complete system stability.
* **Dynamic Health Reporting:** The global capability state is exposed via the root API health router (`/`). If the core system detects that HTTP/3 is disabled or unsupported, the frontend hook intercepts the component mount and masks the UI elements entirely.
* **Idempotency Guard:** Caches and restores the original `default.conf` on every container start, allowing operators to dynamically toggle the `NPM_HTTP3_DISABLED` kill-switch across restarts without container rebuilds.

### 2. Socket `reuseport` Conflict Resolution
* **The Strategy:** Declaring `reuseport` multiple times on identical sockets crashes Nginx. The master QUIC socket is initialized globally as a singleton inside `default.conf`. 
* **UI Sanitization:** Added a strict line-anchored regex sanitizer in `backend/internal/proxy-host.js` (for both create and update operations) that programmatically strips manual `reuseport` declarations from user-submitted `advanced_config` textareas before database commit, preventing user-inflicted Nginx boot failures and cleaning up trailing punctuation cleanly.

### 3. Standards-Compliant API Type Coercion
* Integrated `http3_support` natively into the model's standard `boolFields` lifecycle array. While the backend migration schema uses a storage-efficient `tinyint` format, the data layer automatically handles the type translation so that the API surfaces standard JSON booleans (`true`/`false`), matching the repo's established model design rules line-for-line.

### 4. Shadow-Safe Alt-Svc Header Inheritance
* To satisfy Nginx's header shadowing rules (where nested location block headers completely replace/supersede outer server headers), the `Alt-Svc` header is declared dynamically inside the default `/` location block and custom location templates (`_location.conf`), ensuring reliable QUIC advertisement on the wire without duplicate header noise or parent scoping erasure.
* Enables the correct TLS 1.3 profile required by QUIC while preserving TLS 1.2 compatibility for legacy TCP clients.

---

## 💾 Database Migration & Rollback
* **Up Migration:** Alter Table migration (`20260527000000_http3_support.js`) adds `http3_support` as `tinyint(1)` defaulting to `0` (opt-in).
* **Defensive Rollback:** The `down` migration defensively updates and zeroes all database rows before dropping the column to prevent template rendering errors during the hot reload cycle.

---

## 🎨 Frontend & Internationalization
* Added a clean **Enable HTTP/3 (QUIC)** toggle switch in the SSLOptions modal tab layout.
* Integrated translation keys (`domains.http3-support`) in **all 22 supported languages** to ensure zero internationalization debt across regional installations.

---

## 🧪 E2E Verification & Hardened Test Results

We carried out multiple layers of automated and manual sandbox testing to guarantee absolute deployment safety:

### ✅ 1. Active Nginx Configuration Compile Validation
We verified that the dynamically generated Nginx configurations parse successfully with zero conflicts under the new s6 pipeline:
> **Command:** `docker exec npm-app-1 nginx -t`
> **Output:**
> ```
> nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
> nginx: configuration file /etc/nginx/nginx.conf test is successful
> ```

### ✅ 2. Live HTTP/3 Wire Advertisement Test
We queried the running upgraded container on port 443 over TLS forcing the exact SNI hostname (`beta.lehn.site`) and verified that the dynamic HTTP/3 advertisement header is active and correctly broadcasting:
> **Command:** `curl.exe -I -k --resolve beta.lehn.site:443:127.0.0.1 https://beta.lehn.site/`
> **Output:**
> ```
> HTTP/1.1 404 Not Found
> Server: openresty
> Date: Tue, 26 May 2026 17:58:33 GMT
> Connection: keep-alive
> Strict-Transport-Security: max-age=63072000; preload
> 
> Alt-Svc: h3=":443"; ma=86400
> ```

### ✅ 3. Clean Baseline-to-Feature Upgrade & Data Integrity Verification
We deployed a clean production container running the current official release to initialize a fresh standard database, populated it with mock proxy host data, and then upgraded the container to this feature branch. 

We ran an automated database validation pass to assert structural and data integrity:
> **Command:** `node verify_upgrade.js`
> **Output:**
> ```
> Opening database natively to verify upgrade: .../database.sqlite
> Table schema columns of proxy_host:
>    - id (INTEGER)
>    - created_on (datetime)
>    - domain_names (json)
>    ...
>    - trust_forwarded_proto (tinyint)
>    - http3_support (tinyint)
> 
> http3_support column exists? ✅ YES
> 
> Retrieved migrated Row 1:
>    - id: 1
>    - domain_names: ["test.lehn.site"]
>    - http3_support: 0
>    - http3_support value is 0? ✅ YES (Opt-in default intact)
> 
> =============================================
> 🎉 UPGRADE AND MIGRATION VERIFIED SUCCESSFULLY!
> =============================================
> ```


ricolehn:
Hi,

I have just pushed a comprehensive polish pass to eliminate the final edge-case vulnerabilities and architectural traps before the deep review. 

Here is a quick summary of the hardening updates added in the latest commits:
1. **L4 Stream Collision Protection:** The backend now intercepts and rejects attempts to create UDP streams on port 443 when HTTP/3 is active, preventing fatal socket conflicts that would crash the container.
2. **Robust Multi-Directive Sanitizer:** Upgraded the advanced config pattern to global word boundaries to cleanly strip manual `reuseport` definitions, even if pasted as a compressed single-line multi-statement block.
3. **Frontend Payload Mapping:** Resolved the data serialization bridge inside `ProxyHostModal.tsx` so the UI toggle converts cleanly from Formik state to the API's backend schema.
4. **TLS Compliance Scoping:** Removed strict protocol overrides to respect global compliance standards (like disabling TLS 1.2 environment-wide) through clean native Nginx inheritance rules.

I also ran the full schema validation and linter checks directly inside the official `nginxproxymanager/nginx-full:certbot-node` test runner to guarantee zero style or specification regressions. 

The full end-to-end testing suite is re-enabled and completely green. Ready for architectural review whenever there is time! 🚀

toviszsolt:
@ricolehn
There was an earlier implementation under #5296 that the developer eventually closed. 

A lot of different things were discussed there regarding this topic, and we’ve already looked into some of them together, such as the settings related to QUIC.

If you have a chance, please take a look at #5296 to see if you can find any information that interests you.

Settings related to QUIC:
```
http {
    # [RECOMMENDED]
    # Enable QUIC retry packets to validate client addresses 
    # and prevent IP spoofing attacks
    quic_retry on;
    
    # [SECURITY RISK]
    # Allow 0-RTT (zero round-trip time) early data for faster connection resumption
    # Warning: This may be vulnerable to replay attacks, use with caution
    ssl_early_data on;
    
    # [RECOMMENDED]
    # Enable Generic Segmentation Offloading for QUIC to improve performance
    # by offloading packet segmentation to the network card
    quic_gso on;
    
    # [RECOMMENDED]
    # Set buffer size for HTTP/3 request and response streams
    # Larger values may improve performance for big uploads/downloads
    http3_stream_buffer_size 64k;
}
```

ricolehn:
Hi @toviszsolt,

Thanks for referencing #5296. It's really helpful to see the previous discussions regarding these default settings. 

When preparing this PR, I looked into past attempts to ensure this implementation builds on top of those insights while focusing heavily on long-term stability, safe database migrations, and robust s6-overlay boot guards to prevent any regression.

I've reviewed the parameters you highlighted and just pushed a commit to integrate the recommended stream buffer size:

1. **`http3_stream_buffer_size 64k;`**: Added to `default.conf` to provide a sensible performance baseline for stream chunks out-of-the-box.
2. **`quic_retry on;`**: This is already active in the proxy host templates to handle basic address validation.
3. **`ssl_early_data`**: I left this off by default due to the replay attack risks associated with 0-RTT on non-idempotent requests (like API POST calls). Keeping it disabled by default feels like the safer approach.
4. **`quic_gso on;`**: I opted not to hardcode this because GSO depends heavily on the host's Linux kernel and socket layer. Enforcing it globally might cause Nginx boot failures if the underlying system or Docker environment doesn't support it.

Since Nginx allows these directives to be overridden in the server context, any advanced user who explicitly wants to enable `quic_gso` or experiment with `ssl_early_data` can still add them manually via the Advanced Config tab.

The PR has been updated with the buffer default and the tests are passing. Let me know if anything else needs to be adjusted! 🚀

PR #5627, opened 2026-06-03
Bump Cloudflare version

jemaltz:
Bumps the version of the Certbot plugin for Cloudflare

## Why

Version 2.15.1 fails to load; testing this as a fix.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

### DIFF
--- a/backend/certbot/dns-plugins.json
+++ b/backend/certbot/dns-plugins.json
@@ -77,7 +77,7 @@
 		"full_plugin_name": "dns-cloudflare",
 		"name": "Cloudflare",
 		"package_name": "certbot-dns-cloudflare",
-		"version": "=={{certbot-version}}"
+		"version": "~=5.6.0"
 	},
 	"cloudns": {
 		"credentials": "# Target user ID (see https://www.cloudns.net/api-settings/)\n\tdns_cloudns_auth_id=1234\n\t# Alternatively, one of the following two options can be set:\n\t# dns_cloudns_sub_auth_id=1234\n\t# dns_cloudns_sub_auth_user=foobar\n\n\t# API password\n\tdns_cloudns_auth_password=password1",


jc21:
I'm a bit confused. I use cloudflare dns plugin too and it installs just fine everywhere. Given that Certbot is already v5.6.0 in 2.15.1, I can only assume you're having problems because you use more than just cloudflare in your NPM instance?

jemaltz:
Interesting...

My instance on version 2.14.0 works fine.

I have only one TLD in NPM, and it uses Cloudflare.

On `latest` (2.15.1), I get this error on startup:
```
nginx-proxy-manager  | Traceback (most recent call last):
nginx-proxy-manager  |   File "/opt/certbot/bin/certbot", line 5, in <module>
nginx-proxy-manager  |     from certbot.main import main
nginx-proxy-manager  | ModuleNotFoundError: No module named 'certbot'
nginx-proxy-manager  | s6-rc: warning: unable to start service prepare: command exited 1
nginx-proxy-manager  | /run/s6/basedir/scripts/rc.init: warning: s6-rc failed to properly bring all the services up! Check your logs (in /run/uncaught-logs/current if you have in-container logging) for more information.
```

Because `certbot` fails to load (in a strange way), I dug into the plugin, because of the note on the 2.15.0 release.

Admittedly, I couldn't find an easy way to see which version of the Cloudflare plugin was installed, but I did notice some old versions of the plugin in `/opt/certbot/lib/python3.11/site-packages/` and figured I'd test upgrading as a solution.

Happy to move this to issues if you'd prefer.

jemaltz:
Okay, I figured out the issue. I think I was suffering from [this issue](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3154) at some point, so I copied /opt to a local folder and bind mounted it.

This caused the no-start in the latest version.

Removing /opt as a bind mount results in a short pause during startup, but startup succeeds with the latest version of NPM.

jc21:
Ok great thanks :)

PR #5706, opened 2026-07-12
fix: replace Oracle Cloud Infrastructure DNS plugin 

Djelibeybi:
## Why

The existing Oracle Cloud Infrastructure DNS plugin locks certbot at 1.15.0. Which is bad. This one does not.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

### DIFF
--- a/backend/certbot/dns-plugins.json
+++ b/backend/certbot/dns-plugins.json
@@ -520,12 +520,12 @@
 		"version": "=={{certbot-version}}"
 	},
 	"oci": {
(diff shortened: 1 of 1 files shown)


Djelibeybi:
The Docker image works fine for me, but I suspect I might be the only person using OCI DNS. 😂

PR #5709, opened 2026-07-14
Estonia language translate

siimaarmaa:
- Translate all 258 UI strings in et.json
- Fix and improve HelpDoc/et markdown files
- Align terminology with UI (e.g. puhverserveri host, ümbersuunamise host)
- Fix markdown formatting in Certificates help text

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -6,6 +6,7 @@
 	"license": "MIT",
 	"scripts": {
 		"dev": "vite",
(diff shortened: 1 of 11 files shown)


siimaarmaa (commented):
fix(i18n): fix TypeScript locale map typing for CI build
Type localeOptions tuples explicitly and build a localeCodeByShort map
so resolveLocale returns a string without unsafe casts.

PR #5711, opened 2026-07-14
Fix access list expansion when enabling proxy hosts

### LINKED ISSUE #5710: Access List is not enforced after disabling and re-enabling a Proxy Host
## Checklist

  - Yes
  - Yes
  - Yes

## Describe the bug

After disabling and re-enabling a Proxy Host, the assigned Access List is no longer enforced.

The Nginx Proxy Manager web interface continues to display the Access List as assigned to the Proxy Host, suggesting that the configured protection is still active. However, requests are forwarded directly to the upstream service without applying either the configured IP restrictions or Basic Authentication.

Opening the Proxy Host configuration and simply clicking **Save** immediately restores the expected behavior.

This can unintentionally expose services to the Internet while the UI still indicates that they are protected.

## Nginx Proxy Manager Version

2.15.1

Also reproduced using the current `jc21/nginx-proxy-manager:latest` Docker image.

## Access List configuration

The Access List used for testing was configured as follows:

- Satisfy: Any
- Pass Authentication: Disabled
- Basic Authentication: Enabled (single test user)
- Access rules:
  - Allow: 10.1.0.0/21
  - Allow: fd00::/64
  - Deny: all

## To Reproduce

1. Create a Proxy Host.
2. Assign the Access List described above.
3. Save the Proxy Host.
4. Verify that the Access List is enforced (authentication and IP restrictions work as expected).
5. Disable the Proxy Host.
6. Re-enable the Proxy Host.
7. Access the Proxy Host from an external network.

## Expected behavior

The assigned Access List should remain active after the Proxy Host is disabled and re-enabled.

Basic Authentication and IP restrictions should continue to be enforced until the Access List is explicitly changed or removed.

## Actual behavior

After re-enabling the Proxy Host, the Access List is no longer enforced.

Requests are forwarded directly to the upstream service without requiring authentication or applying the configured IP restrictions, even though the NPM UI still shows the Access List as assigned.

Opening the Proxy Host configuration and saving it again immediately restores the expected behavior.

## Operating System

- Debian
- Docker Compose
- Docker image: `jc21/nginx-proxy-manager:latest`
- Also reproduced with `jc21/nginx-proxy-manager:2.15.1`

## Additional context

This issue appears to have security implications because administrators may believe a service is still protected by an Access List while it is actually publicly accessible.

The behavior is consistently reproducible on my installation.

Opening and saving the Proxy Host immediately restores the expected behavior until the Proxy Host is disabled and re-enabled again.

I can provide logs, screenshots, generated Nginx configuration files, or any additional information required to help reproduce and diagnose the issue.

CarminKern:
## Why

When a disabled Proxy Host is enabled again, the `enable()` code path only expands the `access_list` relation itself.

However, the nginx configuration template also requires the nested `access_list.clients` and `access_list.items` relations to generate the configured IP access rules and Basic Authentication entries.

Because these nested relations are missing, the regenerated nginx configuration does not contain the configured Access List rules after disabling and re-enabling the Proxy Host.

This change loads `access_list.[clients,items]` in the `enable()` path, matching the existing create and update code paths.

Tested by:

- Creating a Proxy Host with an Access List
- Verifying that the Access List is enforced
- Disabling and re-enabling the Proxy Host
- Verifying that the Access List is still enforced afterward
- Building and running the modified Docker image on my server

Fixes #5710

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage


Alexis-Loskoutoff:
Thanks for your PR !
The fix looks good. I've tested the docker image and the bug is fixed.
But 1 thing I'd like to mention : now the password of the access list are shown in clear text in the audit log. You can check in the audit log tab in the UI.

CarminKern:
@Alexis-Loskoutoff  Thanks for pointing this out! I've updated the PR to mask the access list passwords before writing the proxy host enable event to the audit log. The full access list data is still used for the nginx configuration, so the original fix remains intact, while sensitive data is no longer written to the audit log.

Could you please test the updated version when you get a chance?

CarminKern:
@Alexis-Loskoutoff  It looks like Jenkins didn't trigger a new build for my latest commits. Could you please retrigger the PR build?

PR #5715, opened 2026-07-18
Auto-split pasted URLs into scheme/host/port on forward host fields

Grasfer:
## Why

A very common workflow when adding a proxy host is copying a service's URL straight from the browser — e.g. `http://192.168.5.150:8096` — and pasting it into the **Forward Hostname / IP** field. Today the whole string lands in the host field, and the user has to manually delete the scheme, move the port to the port field, and set the scheme dropdown.

With this change, pasting a full URL into the Forward Hostname / IP field (on the proxy host form and on custom locations) parses it and distributes the parts: the scheme dropdown and port field are filled, and only the bare hostname/IP stays in the host field. Pasting a plain hostname or IP behaves exactly as before — the split only triggers when the pasted text parses as a URL with a scheme.

No API or backend changes; frontend-only. Unit tests included for the URL parsing module (`ForwardUrl.test.ts`).

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [x] Test addition or update

## AI Usage

### DIFF
--- a/frontend/src/components/Form/LocationsFields.tsx
+++ b/frontend/src/components/Form/LocationsFields.tsx
@@ -2,9 +2,10 @@ import { IconSettings } from "@tabler/icons-react";
 import CodeEditor from "@uiw/react-textarea-code-editor";
 import cn from "classnames";
 import { useFormikContext } from "formik";
-import { useState } from "react";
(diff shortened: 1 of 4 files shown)


Grasfer:

PR #5716, opened 2026-07-18
Auto-select a matching SSL certificate from typed domains (opt-in secure defaults via NPM_SECURE_DEFAULTS)

Grasfer:
## Why

When adding a new proxy host you type the domain names, then have to hunt through the certificate dropdown for the cert that covers them — error-prone on installs with many certificates, and picking the wrong one silently breaks SSL on the host.

**Certificate auto-pick.** With this change, typing domains on a NEW proxy host auto-selects a certificate covering **all** entered domains, preferring an exact match over a one-label wildcard match. The selection is tracked as auto-picked, so it re-matches or clears as the domains change — but it **never overrides a manual choice**, including an explicit manual "None" (detected via the field's touched state). Editing existing hosts is untouched. As part of this, `SSLCertificateField`'s select becomes a controlled component so programmatic `certificateId` changes render correctly.

**Opt-in secure defaults.** New env var `NPM_SECURE_DEFAULTS` (default **off** = exactly stock behavior), exposed to the frontend via a new `secure_defaults` field in the health endpoint payload. When enabled, new proxy hosts default **Block Common Exploits** and **Websockets Support** on, and **Force SSL** + **HTTP/2 Support** flip on when a certificate is first selected. New hosts only — editing an existing host never changes stored values. Unset, nothing changes for anyone. This part is separable — happy to split it into its own PR if you'd rather take the auto-pick alone.

API surface: one additive, optional field (`secure_defaults`) in the health response, schema updated accordingly. No breaking changes. Unit tests included for the certificate matching logic (`CertificateMatch.test.ts`).

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] API changes
- [ ] Performance improvement
- [x] Test addition or update

## AI Usage

### DIFF
--- a/backend/lib/config.js
+++ b/backend/lib/config.js
@@ -2,15 +2,17 @@ import fs from "node:fs";
 import NodeRSA from "node-rsa";
 import { global as logger } from "../logger.js";
 
-const keysFile               = '/data/keys.json';
-const mysqlEngine            = 'mysql2';
-const postgresEngine         = 'pg';
-const sqliteClientName       = 'better-sqlite3';
+const keysFile = "/data/keys.json";
+const mysqlEngine = "mysql2";
+const postgresEngine = "pg";
+const sqliteClientName = "better-sqlite3";
 
 // Not used for new setups anymore but may exist in legacy setups
-const legacySqliteClientName = 'sqlite3';
+const legacySqliteClientName = "sqlite3";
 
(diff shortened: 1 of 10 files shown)


Grasfer:

rajubepari980-oss (commented):
Off

PR #5717, opened 2026-07-18
Warn when typed domains aren't covered by the selected certificate

Grasfer:
## Why

It's currently easy to save a proxy host with an SSL certificate that doesn't actually cover the typed domain names — nothing in the UI checks that the selected certificate matches. The mistake only surfaces later as browser certificate errors on the live host, which is confusing to debug because the UI showed no problem at save time.

This adds a **non-blocking warning** to the Proxy Host, Redirection Host, and 404 Host modals when one or more typed domains are not covered by the selected certificate (exact or one-label wildcard match). The warning appears under both the domain names field and the SSL certificate selector, since the two live on different tabs and the user may be looking at either one.

Deliberately non-blocking: there are legitimate reasons to save a mismatched pair (cert renewal in flight, DNS not cut over yet), so this informs rather than prevents. The warning stays silent when no certificate is selected ("None"), when requesting a new certificate inline, and while the certificate list is still loading — no false positives during normal flow.

Frontend-only; no API or backend changes. The domain/cert matching logic lives in a new `CertificateMatch` module with unit tests. Note: #5716 adds the same module for certificate auto-selection — whichever lands second rebases trivially.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [x] Test addition or update

## AI Usage

### DIFF
--- a/frontend/src/components/Form/CertificateMismatchWarning.tsx
+++ b/frontend/src/components/Form/CertificateMismatchWarning.tsx
@@ -1,0 +1,30 @@
+import { IconAlertTriangle } from "@tabler/icons-react";
+import { useFormikContext } from "formik";
+import { useCertificates } from "src/hooks";
+import { T } from "src/locale";
+import { uncoveredDomains } from "src/modules/CertificateMatch";
+
+/**
+ * Non-blocking warning shown when the selected certificate does not cover
+ * every typed domain name. Silent for "None" (0), "new" (no domains yet
+ * to compare against) and while the certificate list is loading.
+ */
+export function CertificateMismatchWarning() {
+	const { data: certificates } = useCertificates();
+	const { values }: any = useFormikContext();
+	const certificateId = values?.certificateId;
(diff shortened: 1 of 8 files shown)


Grasfer:

jc21:
Thanks for taking the time on this.

`frontend/src/components/Form/CertificateMismatchWarning.tsx:20` — Custom (uploaded) certificates get false mismatch warnings. For these certificates the backend saves only the certificate's main name (the CN) as its domain list and drops the extra names (the SANs), at `backend/internal/certificate.js:617`.
- Example: an uploaded certificate for `example.com` that also covers `www.example.com` shows "does not cover: www.example.com", even though the host works fine over HTTPS.
- No CN: a certificate with no CN, which is common now, has an empty domain list, so every domain is flagged.
- Fix: skip the warning for custom certificates, or read the extra names when the certificate is uploaded.

PR #3392, opened 2023-12-15
Make auto-renew use built-in renew function

stevecrozz:
As a complete newcomer to this project, this is my attempt to fix certificate auto-renew issues #1916 where manual renew works, but automatic renew does not. This changes proposes we replace the functionality inside automatic renewal with a call to the same function used for manual renewal.

Specifically, I have replaced the certbot 'renew everything' command with a DB-driven one-by-one renewal. Determining which certificates need renewal is a pretty naive time based DB query, and access control is completely bypassed with `Promise.resolve({ permission_visibility: 'all' }` because I assume it isn't needed for a background job.

I'm not sure yet how to best contribute to this project, and I'm sure there are some issues with this so please do guide me as I'm brand new here.

PR #5720, opened 2026-07-19
Fix nginx reload after certificate renewal

### LINKED ISSUE #3803: SSL certificates expired
**Checklist**
  - Yes
  - Yes
  - Yes

**Describe the bug**

The NPM uses a timer to regularly update expired SSL certificates. However, after the certificate is successfully updated, no nginx reload operation has been performed, resulting in nginx still using the old version of the certificate.
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/d3a654b546f1c1b695aa9e2f290e9a72fa355e16/backend/internal/certificate.js#L923-L959

However, if it's about creating new ones, there are operations for reloading in the code.
https://github.com/NginxProxyManager/nginx-proxy-manager/blob/d3a654b546f1c1b695aa9e2f290e9a72fa355e16/backend/internal/certificate.js#L145-L199

2.11.2

**To Reproduce**
1. Setup cert for a host
2. Wait for 60 days
3. Audit log shows that certificate updated
4. Access to the site we add before, it shows 'cert expired'
5. Restart the container or manully reload nginx, the new cert work now

**Expected behavior**

Reload nginx after cert update

**Screenshots**

**Operating System**

Linux

**Additional context**

DoyoDia:
## Why

  A successful Let's Encrypt renewal updates the certificate files on disk, but nginx continues serving the certificate already loaded in memory until it is reloaded.

  The shared certificate renewal path did not reload nginx after Certbot completed. This affects both HTTP and DNS challenge renewals, including Cloudflare.

  This change calls the existing nginx reload function immediately after a successful renewal, allowing the renewed certificate to be served without restarting the container or manually reloading nginx.

  This restores behavior that was lost when auto-renewal was changed to use the shared renewal function in #3392.

  Fixes #3803

  ## Type of Change

  - [ ] Documentation update
  - [ ] Code refactoring
  - [ ] API changes
  - [ ] Performance improvement
  - [ ] Test addition or update

  ## AI Usage

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -904,6 +904,7 @@ const internalCertificate = {
 				: internalCertificate.renewLetsEncryptSsl;
 
 			await renewMethod(certificate);
+			await internalNginx.reload();
 			const certInfo = await internalCertificate.getCertificateInfoFromFile(
 				`${internalCertificate.getLiveCertPath(certificate.id)}/fullchain.pem`,
 			);


mews-se:
Can confirm this on 2.15.1 with http-01 renewals: "Renew now" succeeded for several certificates (certbot renewed them and the UI showed the new expiry dates), but nginx kept serving the old — by then expired — certificates until a manual `nginx -s reload` inside the container. The log shows no "Reloading Nginx" line after the renewals, and `internalCertificate.renew` in the current backend indeed never calls the reload. This patch does exactly what I had to do by hand.

mews-se:
Tested the pr-5720 dev image today. I stubbed out certbot inside the container so a renew just drops a fresh self-signed cert on disk, then compared the serial nginx actually serves on 443 with the one in /etc/letsencrypt/live. On 2.15.1 the disk gets the new serial while nginx keeps serving the old one, exactly what bit me earlier. With this image the served cert switches right after the renew call, no manual reload needed.

Issue #4288, opened 2025-01-10
[BUG] Deleting a certificate without detaching it from the hosts breaks NPM


Grishkaone:
**Describe the bug**
I use NPM on a Rasberry Pi. Data is stored in 2 volumes: one for `/data` and another for `/etc/letsencrypt`.
I have 13 hosts which are all sub-domains of my main domain.
I generate a Let's Encrypt wildcard certificate with a DNS challenge.
All this worked very well.

I recently changed registrar for my domain. Once the migration was completed with the DNS zones up and running, I had to change the method of generating the wildcard certificate (change of API and identifier).

It's not possible to edit this information in NPM. I therefore deleted the certificate, with a view to creating a new one. The deletion was carried out without error. Afterwards, it was impossible to do anything in the back office. I was unlocated and couldn't relocate. I restarted the container, but nothing changed.

There's an error `nginx proxy manager nginx: [emerg] cannot load certificate`: what I understand is that the 13 hosts that are still attached to the certificate are trying to load it without success since it's no longer there. I no longer have access to the BO to detach hosts from the certificate.

To get back in control, here are the steps I took: 
1. Connect to the container console
2. Open `/data/nginx/proxy_host/`
3. Delete all `*.conf` files

Then I was able to reconnect to the BO, the hosts are still there, but no certificates are linked to them.
I could set up my new certificate and attach it to each host one by one. The `*.conf` files were regenerated at this point.

**Nginx Proxy Manager Version**
2.11.3

**To Reproduce**
On a configuration with several hosts attached to a wildcard certificate.
1. Open the **SSL Certificates** tab
2. Click on the menu to the right of the certificate and select **Delete**.
3. Confirm deletion.
NPM becomes unusable.

**Expected behavior**
I can see several solutions.
1) **Prohibit deletion of a certificate attached to an active host¨¨. Add a warning when deleting the certificate to indicate that x hosts are bound to this certificate and not allow deletion until they have been manually detached.
2) **Automate the process**. A message similar to solution 1, but indicating that deletion will automatically detach the certificate from each host beforehand, with a checkbox to indicate agreement.

**Screenshots**
No screenshot, sorry, everything work fine now.

**Operating System**
Linux 6.1.21-v8+ #1642 SMP PREEMPT Mon Apr  3 17:24:16 BST 2023 aarch64 GNU/Linux
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
Debian version : 11.11

[...]

Issue #3465, opened 2024-01-18
Deleting Active SSL Certificate Breaks All Hosts Using It


coanghel:
**Checklist**
- Have you pulled and found the error with `jc21/nginx-proxy-manager:latest` docker image?
  - Yes 
- Are you sure you're not using someone else's docker image?
  - Yes 
- Have you searched for similar issues (both open and closed)?
  - Yes 

**Describe the bug**

Deleting a certificate that is currently in use sends NPM into a reboot loop. Similar to #3420

**Nginx Proxy Manager Version**

2.10.4

**To Reproduce**
1. Configure a proxy host
2. Configure an SSL cert and add it to the above
3. Delete the proxy host
4. NPM will now return "internal error" for any attempted activities (such as swapping hosts to a new certificate); recreating the container presents the below error, where npm-# is the certificate you just deleted.
```
[emerg] cannot load certificate "/etc/letsencrypt/live/npm-3/fullchain.pem": BIO_new_file() failed
```

**Expected behavior**

Either do not allow an actively used certificate to be deleted, or update the nginx.conf files for all hosts using it prior to resolving the delete request.

**Operating System**

Docker running on Ubuntu server

**Additional context**

Upon investigating, I found that the sqlite database reflected the new SSL certificate in the `certificate` table and the hosts I swapped to the new certificate reflected the correct certificate in the `proxy_host` table, however the nginx .conf files in /proxy_host/ folder still had the path of the old certificate in them. Manually editing these config files and then restarting the container allowed it to start up as expected.

zimbres:
Same problem here

flinthamm:
Same problem here

PietroBer:
unfortunately i made the same mistake of deleting 1 ssl certificate and now NPM is in a loop.

GeorgeV220:
How is that not fixed yet??

Cha0yue:
In my case I deleted an active certificate and wanted to create a new certificate with the same domain.
I am using version 2.13.5 and here are the logs when deleting an active certificate:
```
[5/16/2026] [1:01:15 PM] [SSL      ] › ℹ  info      Revoking LetsEncrypt certificates for Cert #1: *.example.com, example.com
[...]

PR #5732, opened 2026-07-22
fix: disable hosts and prevent nginx crash when certificate is deleted

mrbaloghakos:
When a certificate is deleted in NPM, the nginx config files for proxy hosts using that certificate were left referencing cert files that no longer exist on disk. This caused nginx to refuse to start on the next restart with:

```
  nginx: [emerg] cannot load certificate ".../fullchain.pem": BIO_new_file()
  failed (SSL: error:80000002:... No such file or directory)
```

Two complementary fixes:

1. `certificate.delete()` now finds all enabled hosts that reference the deleted certificate, removes their nginx config files, sets `meta.nginx_online=false` with `nginx_err='Certificate was deleted'`, and reloads nginx. Hosts are taken fully offline (rather than silently downgraded to HTTP) so the user sees a clear signal in the UI.

2. A new prepare script (`55-nginx-validate.sh`) runs before nginx starts and scans all host config files for `ssl_certificate` directives. Any config that references a cert file missing from disk is renamed to `.conf.err` so nginx never attempts to load it. This acts as a safety net for scenarios where cert files go missing without NPM knowing (volume recreation, manual deletion, etc.).

Also adds a Cypress test that verifies a proxy host is taken offline when its certificate is deleted.

ℹ️  _Vibecoded using `Claude Sonnet 4.6`_

### Fixes the following issues:
- [Deleting Active SSL Certificate Breaks All Hosts Using It · Issue #3465 · NginxProxyManager/nginx-proxy-manager](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/3465)
- [[BUG] Deleting a certificate without detaching it from the hosts breaks NPM · Issue #4288 · NginxProxyManager/nginx-proxy-manager](https://github.com/NginxProxyManager/nginx-proxy-manager/issues/4288)

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

- [x] AI was used to write this (Claude Sonnet 4.6)

### DIFF
--- a/backend/internal/certificate.js
+++ b/backend/internal/certificate.js
@@ -13,6 +13,10 @@ import error from "../lib/error.js";
 import utils from "../lib/utils.js";
 import { debug, ssl as logger } from "../logger.js";
 import certificateModel from "../models/certificate.js";
+import deadHostModel from "../models/dead_host.js";
+import proxyHostModel from "../models/proxy_host.js";
+import redirectionHostModel from "../models/redirection_host.js";
+import streamModel from "../models/stream.js";
 import tokenModel from "../models/token.js";
 import userModel from "../models/user.js";
 import internalAuditLog from "./audit-log.js";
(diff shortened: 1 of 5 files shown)


jc21:
Thanks for taking the time to fix this. I've found some issues, some for consideration, some need to be resolved.

Consideration:
1. backend/internal/certificate.js:424, files deleted too early: the Let's Encrypt revoke deletes the cert files before the configs of hosts using that cert are removed. The revoke is a network call that can take several seconds, and during it nginx -t fails on the missing files. If another host is saved in that window, it gets marked offline with an unrelated error, or the reload is rejected. Removing the configs before revoking fixes this.
2. backend/internal/certificate.js:468, false delete error: the new internalNginx.reload() runs nginx -t after the cert is already marked deleted, audited and revoked. If that test fails for any reason (for example, some unrelated host's config is broken), the DELETE returns a 500 even though the delete fully happened.
3. backend/internal/certificate.js:430, hosts come back over plain HTTP: affected hosts stay enabled and keep pointing at the deleted cert. Because the cert lookup skips deleted certs, any later config rebuild writes a config with no SSL lines. Updating an attached access list, or saving the host itself, brings it back over plain HTTP, which is the downgrade the PR says it avoids. With an access-list update, the UI also still shows the host as offline with "Certificate was deleted".

Require fixing:
1. docker/rootfs/etc/s6-overlay/s6-rc.d/prepare/55-nginx-validate.sh:28, container can fail to start: the new script renames config files with mv but lacks the is_mounted check that 50-ipv6.sh has. If a user has bind-mounted a single .conf file and its cert is missing, mv fails, set -e stops the prepare stage, and the container won't start.

PR #5742, opened 2026-07-25
fix: trim whitespace from domain names before saving hosts

### LINKED ISSUE #5654: Remove spaces from domain input
**Checklist**
  - **Yes** / No
  - **Yes** / No
  - **Yes** / No

**Describe the bug**
When entering a new host on npm, ex. a domain or an ip, should you put a space before or after said host, it wont load at all and throw a 502 which is highly misleading (as its not related to upstream not responding).
The browser will forcefully remove every space on the domain/hostname so the browser guesses right. 
It's about following the standard and avoiding confusion as someone would instinctively go check for upstream/docker errors.

A javascript check to remove spaces is presumably enough in my opinion.

latest (as per rn)
[v2.15.1](https://github.com/NginxProxyManager/nginx-proxy-manager/releases/tag/v2.15.1)

**To Reproduce**
As mentioned above.

**Expected behavior**
502 Error.

addielaruee:
## Why

Fixes #5654.

When a domain/host is entered with a leading or trailing space (easy to do via copy‑paste), NPM stores the space verbatim. The generated nginx config then contains an invalid `server_name`/upstream, so the host **saves but never loads** — the user sees a misleading `502` that looks like an upstream problem rather than bad input.

### Fix

The `proxy_host`, `dead_host` and `redirection_host` models each sorted `domain_names` in their `$beforeInsert`/`$beforeUpdate` hooks but never trimmed them, and `certificate.js` carried its own near‑identical (also untrimmed) `cleanDomainNames`.

This adds a single shared `cleanDomainNames` helper in `backend/lib/helpers.js` that:

- drops `null`/`undefined` entries,
- **trims surrounding whitespace**,
- removes any entry left empty after trimming,
- sorts,

and uses it across all four models, replacing the ad‑hoc per‑model `.sort()` and removing the duplicated copy in `certificate.js`. Trimming now applies consistently to every host type (and certificates) at the persistence layer.

Behaviour is a superset of the previous logic (which already filtered `null` and sorted), so existing well‑formed input is unaffected.

### Verification

- `biome lint` passes on all five changed files.
- Unit-style checks of the helper:

  | Input | Output |
  |---|---|
  | `[" example.com ", "b.com"]` | `["b.com", "example.com"]` |
  | `["  spaced.test  "]` | `["spaced.test"]` |
  | `["z.com", "a.com", ""]` | `["a.com", "z.com"]` |
  | `["keep.com", null, "  ", "  x.io"]` | `["keep.com", "x.io"]` |
  | `undefined` | `[]` |

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

### DIFF
--- a/backend/lib/helpers.js
+++ b/backend/lib/helpers.js
@@ -55,4 +55,24 @@ const convertBoolFieldsToInt = (obj, fields) => {
  */
 const castJsonIfNeed = (colName) => (isPostgres() ? ref(colName).castText() : colName);
 
-export { parseDatePeriod, convertIntFieldsToBool, convertBoolFieldsToInt, castJsonIfNeed };
+/**
+ * Normalizes a domain_names array before it is persisted: drops null/undefined
+ * entries, trims surrounding whitespace, removes any entry that is left empty,
+ * and sorts. A leading or trailing space in a hostname produces an invalid
+ * nginx server_name/upstream, so it must be stripped before the config is built.
+ *
+ * @param   {Array}  domainNames
+ * @returns {Array}
+ */
+const cleanDomainNames = (domainNames) => {
+	if (typeof domainNames === "undefined") {
+		return [];
+	}
+	return domainNames
+		.filter((name) => name != null)
+		.map((name) => name.trim())
+		.filter((name) => name.length > 0)
+		.sort();
+};
+
+export { parseDatePeriod, convertIntFieldsToBool, convertBoolFieldsToInt, castJsonIfNeed, cleanDomainNames };
--- a/backend/models/certificate.js
+++ b/backend/models/certificate.js
@@ -3,7 +3,7 @@
 
 import { Model } from "objection";
 import db from "../db.js";
-import { convertBoolFieldsToInt, convertIntFieldsToBool } from "../lib/helpers.js";
+import { cleanDomainNames, convertBoolFieldsToInt, convertIntFieldsToBool } from "../lib/helpers.js";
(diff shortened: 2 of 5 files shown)


jc21:
Thanks for taking the time to fix this.

I found 2 issues. Both come from the trimming being done too late: it happens in the model save hooks, after the duplicate-host check and schema validation have already run on the untrimmed names.

- `backend/models/proxy_host.js:42` — Duplicate hosts can get through. The "already in use" check (`internalHost.isHostnameTaken`) runs before the model trims. So if host A already has `example.com` and someone creates host B with `example.com` plus a trailing tab, the check finds no match. Host B is then saved as plain `example.com`, two hosts share one `server_name`, and nginx serves only one of them. The same applies to `dead_host.js` and `redirection_host.js`.
- `backend/lib/helpers.js:68` — Trimming can undo what the schema guarantees. The schema requires at least one name and no repeats, but it checks before trimming. So `['a.com', 'a.com\t']` is saved as `['a.com', 'a.com']`. And `['\t']` is saved as an empty list, which renders an empty `server_name ;`, so the nginx config fails to load and the host is broken again.

The regex pattern for `domain_names` in `backend/schema/common.json` already rejects a literal space. So the copy-pasted leading or trailing spaces from issue #5654 may already be refused by the API before these hooks run, and the new trimming would only ever act on other whitespace such as tabs. That's worth checking against the actual reproduction before merging.

PR #5790, opened 2026-08-21
Switched to local QR code generation to prevent leaking 2FA keys to third-party QR services.

vgoer:
## Why

Background
When 2FA was enabled, the original implementation embedded the `otpauth://` link (containing the TOTP secret) into the `src` attribute of an `` tag and sent it to the third-party QR code service `api.qrserver.com` to generate the QR code. The browser automatically initiated this request without user interaction; consequently, the TOTP secret for every user enabling 2FA was transmitted in plaintext to an external service.

Risks
1. TOTP secrets were exposed to a third-party service; the provider could store them indefinitely and replay 2FA codes, rendering 2FA ineffective.
2. Dependency on external service availability: users could not complete 2FA setup if `qrserver.com` experienced an outage.
3. The request also transmitted account information, creating a potential privacy leak.

Changes
1. Removed the dependency on `api.qrserver.com`.
2. Integrated the `qrcode` library to generate QR codes locally within the browser (using data URIs), ensuring the TOTP secret never leaves the browser.
3. Added new dependencies: `qrcode ^1.5.4` and `@types/qrcode ^1.5.6`.

## Type of Change

- [ ] Documentation update
- [ x] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

- [x ] AI was used to review this


Alexis-Loskoutoff:
Thanks for your PR!
One thing that would help with the review: this PR also includes your commit about permissions fix. Could you please keep only the commits related to the QR code generation ?
Another thing I'd like to highlight is that the qrcode package doesn't seem to be actively maintained, with the last commit dating back about two years. It might be worth considering a more actively maintained alternative before introducing this dependency.

vgoer:
I apologize, sir. This is my first time contributing code to another repository, so I wasn't fully aware of the conventions. I will close this pull request, refine the code based on your suggestions, and then submit the correct code to the repository in accordance with the standards.

PR #5793, opened 2026-08-21
Insufficient private key permissions led to a leak.

vgoer:
## Why
Background
When the RSA private key used for JWT signing was written to `/data/keys.json`, no specific file permissions were set, resulting in the system default of `0644` (readable/writable by the owner, readable by others).

Risks
1. `/data` is a bind mount from the host machine; the private key file is effectively exposed to the host's file system.
2. Any local user or process within a container (e.g., a compromised service) with access to this directory can read the private key.
3. Possession of the private key allows for the offline forgery of JWTs for any identity (e.g., an administrator with `attrs.id=1`). This enables a complete system takeover without requiring account credentials and bypasses 2FA (as the forged token circumvents the standard login flow).

Changes
1. `backend/lib/config.js`: Explicitly specified `{ mode: 0o600 }` in the `writeFileSync` call within `generateKeys()`.
2. Existing instances require a manual `chmod 600 /data/keys.json` (this step has already been executed as part of the current fix).

Impact
- The private key is readable only by the owner (root/container service process); access is denied to all other users.
- The attack vector for forging JWTs is completely eliminated.
- No functional impact: JWT issuance and verification logic remain unchanged; only file permissions have been tightened.

Verification
- Before fix: `nobody` user could read the private key → forged token used to call `/api/users` returned HTTP 200 (full takeover).
- After fix: `nobody` user attempting to read the private key received "Permission denied"; service operated normally (HTTP 200 on port 81).

## Type of Change

- [x ] Bug fix (non-breaking change that fixes an issue)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] API changes
- [ ] Performance improvement
- [ ] Test addition or update

## AI Usage

- [ x] AI was used to review this


Alexis-Loskoutoff:
Thanks for your PR!
One thing that would help with the review: this PR also includes your commit about local QR code generation.
Could you please keep only the commits related to the permissions issue ?
Have you checked that the fix work when PUID and PGID as set as environment variables ?

vgoer:
I apologize, sir. This is my first time contributing code to another repository, so I wasn't fully aware of the conventions. I will close this pull request, refine the code based on your suggestions, and then submit the correct code to the repository in accordance with the standards.

PR #5797, opened 2026-08-22
Add LDAP, SAML and OAuth authentication providers

timothestoifl24:
## Why

There are long-standing requests for LDAP sign in, and separate ones for SAML
and other identity providers. Rather than bolting on a single protocol, this
adds a general **authentication provider** concept so the same plumbing covers
all three and leaves room for more.

It also picks up the work from #5345 (thanks @Wadera) and answers the question
@adamoutler raised there: providers are not limited to the admin dashboard, they
can protect proxied sites through access lists too.

| Type | Protocol | On the login screen |
| --- | --- | --- |
| LDAP | LDAP / LDAPS, optionally StartTLS | the usual username and password form |
| SAML | SAML 2.0 | a "Continue with …" button |
| OAuth | OAuth 2.0 / OpenID Connect | a "Continue with …" button |

LDAP deliberately reuses the existing login form, so directory users type their
username or email like everyone else and that page does not change for them.
Providers are configured under **Users → Authentication Providers**, or entirely
through `AUTH_*` environment variables for container deployments.

## What is in it

**Signing in.** Per-provider opt-in for creating accounts on first sign in;
otherwise an administrator creates the account and it is matched by email. An
optional admin group grants the `admin` role on every sign in and revokes it when
somebody leaves the group. Local password sign in can be switched off once a
provider is enabled, guarded so an instance cannot be left with no way in, with
`AUTH_DISABLE_LOCAL` overriding the stored setting either way.

**Stable directory identifiers.** Accounts are keyed on `objectGUID` (Active
Directory) or `entryUUID` (OpenLDAP) rather than the DN, because a DN changes
when somebody is renamed or moved between organisational units and used to
strand their account. `auth.identifier` carries a unique index per provider so
two accounts cannot bind to the same directory object.

**Directory sync**, per provider. Walks the directory on a schedule so accounts
exist before anyone signs in and group-driven roles stay current without waiting
for a login. Optionally disables accounts whose entry has gone away, guarded so
the last remaining administrator is never disabled and a run returning nothing
at all disables nobody. Paged search handles directories past the server's
result cap.

**Access lists** gain a Providers tab, so a proxied site can accept directory
accounts, optionally restricted to named groups. Lists with no providers keep
the existing htpasswd path untouched; only those that opt in switch to an
`auth_request` subrequest against `GET /access-lists/{id}/verify`. Decisions are
cached (5 minutes, refusals 30 seconds, dropped when the list is saved) because
that check sits in front of every request. An unreachable directory refuses, so
a protected site does not fall open. `X-Auth-User` and `X-Auth-Email` are passed
to the proxied application.

**Testing a provider before committing to it.** A Test connection button sits
inside the settings dialog next to the fields it exercises, so the connection can
be confirmed before the rest is filled in. LDAP protocol errors are translated,
so a wrong bind password reads "Invalid credentials — check the bind DN and
password" rather than ldapts' bare `Code: 0x31`.

**An Auth column** in the Users list shows where each account signs in from,
naming the actual provider rather than a generic badge.

### API changes

Additive only; no existing endpoint changes shape. New: `GET /api/auth/providers`
(unauthenticated, names and types only), `POST /api/auth/exchange`, the redirect
endpoints under `/api/auth/{id}/…`, admin CRUD under `/api/auth-providers` with
`test`, `test-credentials` and `sync`, and `GET /api/access-lists/{id}/verify`
(unauthenticated by design — it checks a site visitor's credentials, not an
administrator's). All have OpenAPI definitions and `validate-schema` passes.

## Testing

Verified end to end against real identity providers — OpenLDAP,
`mock-oauth2-server` and simpleSAMLphp — not mocks:

- LDAP login by uid and by email, auto-provisioning, group → admin mapping
  **including revocation**, and an account surviving a move between OUs
- Full OAuth code flow; replayed codes, forged `state` and spent `state` rejected
- Full SAML flow; **assertion tampering, corrupted signatures and stripped
  signature blocks all rejected**
- Directory sync creating an account with no login, then disabling it when the
  entry was deleted
- Access list verification: correct password in group passes, correct password
  outside the group refused, wrong password, unknown user and empty password all
  refused, local entries still work
- Generated nginx config checked with `nginx -t`; NPM's nginx is built
  `--with-http_auth_request_module`
- Browser pass over the UI throughout

102 backend tests added (vitest). Backend lint and schema validation pass; the
frontend typechecks, lints and builds.

`docker/docker-compose.auth-dev.yml` reproduces the whole stack with one command,
all three providers pre-wired, credentials in the file header.

## Known issues — please read before merging

I reviewed my own work and did not fix everything. These are real and outstanding:

1. **Unverified email allows account takeover** (`lib/auth/provision.js`).
   An identity is linked to an existing local account on an email match with no
   check that the provider verified it, and that happens *before* the
   auto-create gate — so turning off "Create users on first sign in" does not
   prevent it. With a provider where users control their own email, setting it to
   an administrator's address takes over that account. Needs an `email_verified`
   check for OIDC and an explicit opt-in for email-based linking.
2. **SAML assertions are replayable** within their validity window
   (`lib/auth/saml.js`). `validateInResponseTo: "never"` with no replay cache;
   `RelayState` is single use but an attacker can mint a fresh one at will.
3. **Attacker-controlled text on the login page** (`internal/auth.js`). The OAuth
   callback reflects the IdP-supplied `error_description` before any `state`
   validation. React escapes it so it is not XSS, but it puts arbitrary text in a
   security-relevant banner on a trusted origin.

Dependency audit: no advisory in either package traces to a dependency this
branch adds. The 49 pre-existing backend advisories all come from `sqlite3`,
`swagger-parser`, `proxy-agent`, `ajv`, `objection`, `nodemon`, `archiver`,
`express` and `liquidjs`, and want a separate bump pass.

## Notes for reviewers

- Kept `ldapts` rather than the `ldapjs` used in #5345: ldapjs was last released
  in December 2023 and its repository is archived, while ldapts covers paging and
  binary attributes natively and pulls one transitive dependency.
- Used vitest rather than #5345's jest, to match the frontend and avoid
  `--experimental-vm-modules`.
- New English locale strings only. The other locales are already out of sync
  (`de` has 218 keys against `en`'s 361), so I followed the existing convention
  rather than committing machine translations.
- Two pre-existing bugs found along the way and deliberately **not** fixed here,
  both unrelated to this work: `internalUser.create` reassigns `user` to the
  inserted auth row so the permissions row gets the wrong id, and
  `lib/access.js:274` constructs `PermissionError` without `new`, which makes
  every authorization failure surface as 404 instead of 403.
- `twoFactor.isEnabled()` **is** fixed here, because it threw for any user
  without a local password row and would have rejected every external login.

## Type of Change

- [x] Documentation update
- [ ] Code refactoring
- [x] API changes
- [ ] Performance improvement
- [x] Test addition or update

## AI Usage


adamoutler:
I'm excited to see this in action. However im in a hotel while renovating both the house I'm selling and moving into. Only one of my racks is operational, so I don't have infrastructure emplaced to test this. It may be a couple of weeks. 

Access list providers is an awesome feature!  Thank you!

Please don't hold up this PR on my account, but I'd like to ask some questions based on the description.

For edge environments with constrained memory and slow I/O (e.g., running on embedded boards or flash storage), can the scheduled directory sync daemon be completely disabled per-provider in favor of pure just-in-time (JIT) authentication/binding? If disabled, does JIT auto-provisioning still respect group-to-role mappings without persisting the entire LDAP directory tree locally?

How does the 'last remaining administrator' guard compute admin quorum during a sync pass? Specifically:
- Does it count only enabled local accounts, or does it also count federated accounts whose admin status is derived dynamically via group claims?
- If AUTH_DISABLE_LOCAL=true and the remote LDAP server becomes unreachable or drops the admin group mapping, how does the system prevent a total administrative lockout?

The email-matching takeover is a critical risk for deployments using public/semi-public OAuth providers. Rather than carrying this as a known issue, can we gate account linking behind an explicit configuration toggle (`link_by_email: false` by default) and enforce `email_verified: true` for OIDC?

timothestoifl24:
Thanks — no rush, and good luck with the move. All three are fixed in 58209f83. Two of your questions found real bugs.

**Sync on small hardware.** Sync is per-provider and off by default. When it's off, no timer runs at all. The directory is never copied locally either way — the only rows written are a user and a link, for the people who actually sign in. Group-to-role mapping happens at sign-in, so **Administrator group** works exactly the same with sync off.

**Admin quorum.** You were right to ask. It counted anyone who *held* the admin role, local or federated. But holding the role isn't the same as being able to sign in. With `AUTH_DISABLE_LOCAL=true`, an admin whose only credential is a password can't get in — and the old check counted them, then deleted the one person who still could. It now checks whether the other admin could actually reach the login screen.

Separately, and worse: if the LDAP group search failed, we treated it as "this user is in no groups". With an admin group configured, that means "not an admin" — so a brief LDAP hiccup quietly demoted every administrator who signed in during it. A failed lookup now leaves roles untouched.

On lockout: you can no longer switch off local sign in until an admin has actually signed in through a provider once. Removing the last provider switches local sign in back on. And if `AUTH_DISABLE_LOCAL` is what's blocking you, we can't override it from inside, so it logs a clear error telling you to unset it.

One thing I can't fix: if your IdP goes down while local sign in is off, nobody can get in until it's back or you restart with `AUTH_DISABLE_LOCAL=false`. That's the trade you make by turning the password form off. It's in the docs now.

**Email matching.** Done as you suggested. `link_by_email` is `false` by default everywhere, and OAuth also requires `email_verified: true` — a missing claim counts as a no, not just a false one. If linking is off and the address already belongs to an account, the sign in is refused with an explanation rather than silently taking the account over.

Heads up if you're already testing this branch: **linking by email was the only reason SAML worked.** The stable ID we stored was the `NameID`, but simpleSAMLphp (and others) issue a new one every session, so every login looked like a new person and the email match quietly covered for it. Once I closed the gap, the second SAML login broke. A transient `NameID` now falls back to the email address, or an attribute you choose. A persistent `NameID` is used as before.

Two smaller fixes went in too: a captured SAML assertion can no longer be replayed (which means IdP-initiated sign in isn't supported — it's in the docs), and the OAuth callback no longer prints the provider's error text onto the login page.

I tested against a real stack, not just unit tests: LDAP, SAML and OAuth all still sign in, and two SAML logins in a row now land on one account instead of two. 126 backend tests pass.

timothestoifl24:
Here are some screenshots of what this looks like 😉

**Login page**

Providers that sign in by redirect get a button each. The password form stays underneath, and disappears entirely if you turn local sign in off. LDAP has no button — it uses the normal username and password form, so directory users sign in the same way everyone else does.

example_startpage

**Users**

The list has an **Auth** column showing where each account comes from: `Local` for a password, or the provider's name. Accounts created by a provider hold no password of their own, so it's useful to see at a glance which is which — especially before removing a provider.

example_users

**Authentication Providers**

A second tab under Users. Each row shows the type, whether it's enabled, and how it provisions people — whether accounts are created on first sign in, and how often the directory is synced. Providers configured through environment variables are marked **From environment** and are read-only here, since the container's config owns them.

**Test** checks the connection without saving anything, so you can confirm a server and bind account work before filling in the rest. **Allow email and password sign in** is the switch that turns off the password form.

example_auth_providers

**Provider configuration**

Fields depend on the type.  Secrets are write-only: they're never sent back to the browser, so an existing value shows as set rather than being displayed, and leaving the field blank keeps it.

***OAuth***
example_oauth_provider_newconfig

***SAML***
example_saml_provider_newconfig

***LDAP***
example_ldap_provider_newconfig

**Access lists**

Request from #5345. An access list gets a **Providers** tab where you tick the directories it should accept. Visitors are still prompted for a username and password exactly as before — those credentials are just checked against the directory instead of only the list.

**Restrict to groups** narrows it further, so a proxied site can be limited to one group rather than everyone in the directory. Leave it empty to accept anyone the directory authenticates. Usernames typed on the Authorizations tab keep working either way, so an existing list carries on unchanged and you can mix the two.

Only LDAP appears here. SAML and OAuth sign people in by redirecting a browser, which can't happen inside the subrequest nginx makes to check a proxied request.

example_accesslist_providers

adamoutler:
Awesome!  I am reading and since I'm mobile, maybe I'm missing something.

1. 
> Fields depend on the type. This one is .

What is this one?
2. LDAP picture link is invalid when clicked. 

3. What happens if I have a built-in account which is linked, and then create an OAUTH/LDAP, login,  and then disable the associated OAUTH or LDAP account?

timothestoifl24:
1. I have updated my previous comment because the formatting via GitHub CLI apparently didn't go through as planned 😅
2. I updated the picture link, please try again
3. The short answer, the local password keeps working. Linking adds a way in, it doesn't replace the one that was already there.

---
I ran exactly your scenario: built-in account with a password, then linked to LDAP by signing in through it:

| Step | Password login | LDAP login |
| --- | --- | --- |
| Both set up | works | works |
| Account removed/disabled in the directory | **still works** | refused |
| After a sync with *Disable accounts that leave the directory* | refused | refused |

So disabling someone upstream closes the SSO door but not the password door.
Only two things actually revoke access: disabling the account here on the Users screen or letting directory sync do it for you which is off by default and only disables accounts belonging to that provider.

I hope this answers your question 😉

PR #5836, opened 2026-09-06
Invalidate tokens issued before a password change

### LINKED ISSUE #5833: Session remains active after username/password change
There appears to be an authentication/session management issue where an existing user session remains active after the user's credentials are changed from another browser/session.

For example:

Log in to Nginx Proxy Manager as User A in Browser 1.
Keep Browser 1 logged in.
Open Browser 2 and log in using the same User A account.
From Browser 2, change User A's username/email and password.
Browser 1 is still logged in and can continue using the application without being asked to authenticate again.
The session in Browser 1 remains valid even though the user's credentials have been changed.
Expected behavior

Changing a user's password and/or username/email should invalidate existing authenticated sessions for that user, or otherwise require re-authentication.

For example, after changing the password from Browser 2, Browser 1 should receive an authentication/session error and be redirected to the login page.

A configurable session timeout would also be useful so that authenticated sessions cannot remain valid indefinitely.

Actual behavior

The existing session in Browser 1 remains active after the user's password and username/email have been changed from another browser.

The user is not prompted to log in again, and the existing session continues to have access to the account.

Security concern

This could be problematic if an account is compromised or a user's credentials are changed as part of an account recovery/security action.

For example:

An administrator changes a user's password because the account may have been compromised.
The previous session/token may still be valid.
Anyone who has access to that previous browser/session can potentially continue accessing the account despite the password having been changed.

Changing the password should ideally revoke previously issued sessions/tokens.

jmrplens:
## Why

Fixes #5833.

Changing a password does not end the sessions the old password opened. Tokens are stateless JWTs with a one day expiry, so the browser that logged in before the change keeps working until that expiry runs out. The issue describes the case where this matters: an administrator resets a password because the account may be compromised, and whoever is already inside stays inside for up to a day.

Measured on `jc21/nginx-proxy-manager:latest`, driving the API directly:

| | before | after |
|---|---|---|
| `POST /api/tokens` with the old password | 400 | 400 |
| `GET /api/users/me` with a token minted before the change | **200** | **401** |
| `GET /api/users/me` with a token minted after the change | 200 | 200 |

The old password stops working immediately, which is why the second row is the whole bug: the credential is gone but the session it opened is not.

**How.** The `auth` row already records when the password last changed, because `$beforeUpdate` stamps `modified_on` on the same patch that re-hashes the secret, so nothing new has to be stored and there is no migration. `Access.init()` already loads the user on every authenticated request; it now reads that row alongside it and refuses a token whose `iat` is older.

Both sides are compared as whole seconds, which is all `iat` carries. Postgres stores `modified_on` to the microsecond, and comparing in milliseconds locked out every token minted in the same second as the row it was measured against, a new user's very first token included. That failure only shows up on Postgres: SQLite's `datetime('now','localtime')` has whole-second resolution, so the same code passed there.

**Why 401 and not 403.** `frontend/src/api/backend/base.ts` clears the session and reloads on 401 and on nothing else, so the browser holding the dead token lands on the login page, which is what the issue asks for. Anything else leaves it sitting on a page full of errors until the token expires on its own. `can()` wraps every failure from `init()` into `PermissionError`, so it now lets this one error through unwrapped. Worth calling out as an API-visible change: a script holding a token across a password change now gets 401 where it used to get 200.

**What it deliberately does not do.** Only the password revokes. A user row changing does not, since that would log people out when an administrator renames them or edits their permissions, and the identity that was rotated is the password. A user with no `password` auth row, which is what an account authenticating through an external provider looks like, is not affected at all. The one second resolution means a token minted in the same second as the change survives it.

## Verification

Their own suite, run against a container built from `jc21/nginx-proxy-manager:latest` with the patch copied in:

- `Users.cy.js` 4 passing, including the new test. Against the unpatched backend that test fails with `expected { Object (id, created_on, ...) } to have property 'error'`, which is the stale token returning the user instead of an error.
- `Settings.cy.js` 7 passing, `Health.cy.js` 3 passing, `Dashboard.cy.js` 1 passing.
- `ProxyHosts.cy.js` fails identically with and without the patch: it needs the dev stack's DNS aliases and certificate provisioning, which this container does not have.

Twelve API checks run by hand on **both SQLite and Postgres**, green on both: normal traffic on `/users`, `/users/me`, `/nginx/proxy-hosts`, `/settings`, `/audit-log`, `/reports/hosts` and `/nginx/certificates`; a newly created user's first token working immediately; a rename not invalidating anything; an administrator changing another user's password killing that user's token while keeping their own; and that user logging back in with the new password.

`biome check` is clean on both changed backend files.

## Type of Change

- [x] API changes
- [x] Test addition or update

## AI Usage

### DIFF
--- a/backend/internal/user.js
+++ b/backend/internal/user.js
@@ -388,11 +388,21 @@ const internalUser = {
 					.andWhere("type", data.type)
 					.first()
 					.then((existing_auth) => {
+						// Stamped here rather than read off modified_on, because it is compared against a
+						// token's `iat` and the two only line up when the same clock writes both. The
+						// database clock is a different one: with the app on one timezone and the database
+						// on another, its timestamps come back hours away from where Node thinks it is.
+						const password_changed_at = Math.floor(Date.now() / 1000);
+
 						if (existing_auth) {
 							// patch
+							const meta = existing_auth.meta || {};
+							meta.password_changed_at = password_changed_at;
+
 							return authModel.query().where("user_id", user.id).andWhere("type", data.type).patch({
 								type: data.type, // This is required for the model to encrypt on save
 								secret: data.secret,
+								meta,
 							});
 						}
 						// insert
@@ -400,7 +410,7 @@ const internalUser = {
 							user_id: user.id,
 							type: data.type,
 							secret: data.secret,
-							meta: {},
+							meta: { password_changed_at },
 						});
 					})
 					.then(() => {
--- a/backend/lib/access.js
+++ b/backend/lib/access.js
@@ -12,6 +12,7 @@ import { fileURLToPath } from "node:url";
 import Ajv from "ajv/dist/2020.js";
 import _ from "lodash";
 import { access as logger } from "../logger.js";
+import authModel from "../models/auth.js";
 import proxyHostModel from "../models/proxy_host.js";
 import TokenModel from "../models/token.js";
 import userModel from "../models/user.js";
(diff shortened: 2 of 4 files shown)


jmrplens:
Your CI was right and I was wrong, so thank you for it: `e4585ac` fixes a real hole in the first commit rather than the test that caught it.

The check compared a token's `iat`, which is UTC seconds from Node, against `auth.modified_on`, which the driver hands back interpreted in the app's timezone. Your CI runs the app on `Australia/Brisbane` and the database container on UTC, so that column comes back ten hours in the past and the token always looks newer than the change. It passed on SQLite, where nothing converts, and silently never fired on the stack where the two clocks disagree. I had tested SQLite and Postgres with matching timezones, which is exactly the configuration that hides it.

The moment is now recorded in `auth.meta.password_changed_at`, written by `setPassword` with the same `Date.now()` that mints `iat`. One clock on both sides, same unit, and no timestamp parsing at all: the `Date` versus local-string branch is gone, and so is the whole-second flooring that Postgres microseconds had forced. Rows written before this carry no marker and revoke nothing until their next password change, which is the safe direction to be wrong in.

Re-verified by reproducing your CI's condition locally, app container on `Australia/Brisbane` against a database container on UTC, with the clock skew confirmed inside the containers (`04:11 AEST` in the app, `18:11 UTC` in MySQL):

| | SQLite | MySQL | Postgres |
|---|---|---|---|
| 12 API checks, including a stale token after an admin resets another user's password | pass | pass | pass |
| `Users.cy.js`, 4 tests | pass | pass | pass |

The Cypress run on MySQL is the one that matters here, since that is the stack your CI failed on.

